PR #10, opened 2023-08-29
MGMT-15538: Add an annotation to edited k8s resources

carbonin:
The annotation has a string value that is a serialized json object with
its keys being JSON pointers of fields in the annotated resource and
values being a list of PEM bundle indices that were edited (or "N/A" if
the field was not PEM-encoded).

For example:

```
metadata:
  annotations:
    ...
    recert-edited: '{"/spec/config/storage/files/15/contents/source":["1","2","6","5","3","0","4"],"/spec/config/storage/files/21/contents/source":["0"]}'
    ...
```

This annotation is purely informational and is not used programmatically
by recert itself. It's used to help troubleshooters notice this is not a
"natural" cryptographic object, but instead one that was manipulated by
recert.

Resolves https://issues.redhat.com/browse/MGMT-15538

### DIFF
--- a/src/file_utils.rs
+++ b/src/file_utils.rs
@@ -104,3 +104,67 @@ pub(crate) fn decode_resource_data_entry(yaml_location: &YamlLocation, value_at_
     }
     .clone())
 }
+
+/// Annotates a kubernetes resources to indicate that it has been edited by recert. The
+/// annotation has a string value that is a serialized json object with its keys being JSON
+/// pointers of fields in this resource edited by recert and values being a list of PEM bundle
+/// indices of cryptographic objects that were edited in that field (or "N/A" if the field was not
+/// PEM-encoded).
+///
+/// For example:
+///
+/// metadata:
+///   annotations:
+///     ...
+///     recert-edited: '{"/spec/config/storage/files/15/contents/source":["1","2","6","5","3","0","4"],"/spec/config/storage/files/21/contents/source":["0"]}'
+///     ...
+///
+/// This annotation is purely informational and is not used programmatically by recert itself. It's
+/// used to help troubleshooters notice this is not a "natural" cryptographic object, but instead
+/// one that was manipulated by recert.
+pub(crate) fn add_recert_edited_annotation(resource: &mut Value, yaml_location: &YamlLocation) -> Result<()> {
+    if resource.pointer_mut("/metadata/annotations").is_none() {
+        resource
+            .pointer_mut("/metadata")
+            .context("metadata must exist")?
+            .as_object_mut()
+            .context("metadata must be an object")?
+            .insert(String::from("annotations"), Value::Object(serde_json::Map::new()));
+    }
+
+    let current_value_string = match resource.pointer_mut("/metadata/annotations/recert-edited") {
+        Some(annotation_data) => annotation_data.as_str().context("recert annotation data must be a string")?,
+        None => "{}",
+    };
+
+    let mut annotation_value: Value = serde_json::from_str(current_value_string).context("parsing recert annotation json")?;
+
+    let edited_index = Value::String(match &yaml_location.value {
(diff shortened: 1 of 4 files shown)


carbonin:
@omertuc this should be good to review

src/file_utils.rs
omertuc, line 1: I think `k8s_etcd.rs` might be a better location for this function

src/file_utils.rs
```diff
@@ -104,3 +104,59 @@ pub(crate) fn decode_resource_data_entry(yaml_location: &YamlLocation, value_at_
     }
     .clone())
 }
+
+pub(crate) fn add_annotation(resource: &mut Value, yaml_location: &YamlLocation) -> Result<()> {
```
omertuc, line 108: nit: function name feels a bit too general

```suggestion
pub(crate) fn add_recert_edited_annotation(resource: &mut Value, yaml_location: &YamlLocation) -> Result<()> {
```

omertuc, line 108: I find this function to be hard to understand without background / an example

```suggestion
/// Adds annotation to kubernetes resources to indicate that they have been edited by recert. The
/// annotation has a string value that is a serialized json object with its keys being JSON
/// pointers of fields in this resource edited by recert and values being a list of PEM bundle
/// indices of cryptographic objects that were edited in that field (or "N/A" if the field was not
/// PEM-encoded).
///
/// For example:
///
/// metadata:
///   annotations:
///     ...
///     recert-edited: '{"/spec/config/storage/files/15/contents/source":["1","2","6","5","3","0","4"],"/spec/config/storage/files/21/contents/source":["0"]}'
///     ...
///
/// This annotation is purely informational and is not used programmatically by recert itself. It's
/// used to help troubleshooters notice this is not a "natural" cryptographic object, but instead
/// one that was manipulated by recert.
pub(crate) fn add_annotation(resource: &mut Value, yaml_location: &YamlLocation) -> Result<()> {
```

src/file_utils.rs
```diff
@@ -104,3 +104,59 @@ pub(crate) fn decode_resource_data_entry(yaml_location: &YamlLocation, value_at_
...
+            Value::String(String::from("N/A"))
+        },
+    };
+
+    match annotation_value.get_mut(&yaml_location.json_pointer) {
+        Some(locations) => {
+            locations
+                .as_array_mut()
+                .context("locations must be an array")?
+                .push(edited_index);
```
omertuc, line 144: Can we sort the indices for a bit of extra determinism?

omertuc (changes requested):
Looks good overall, some nitpicks. 

A few other points:

* We also need to think about filesystem (as opposed to etcd) YAMLs (`/etc/machine-config-daemon/currentconfig` is a machineconfig that you can find on the filesystem, and there are a bunch of `kubeconfig`s under `/etc/kubernetes` and `/var/lib/kubelet`

* Please state the JIRA ticket in the title / commit message

src/file_utils.rs
carbonin: Sure, but given that I went with strings (to accommodate non-PEM cases) it might sort strangely with indexes over 9.

carbonin: So I looked into this and because the elements at this point are `Values` (which don't have `Ord`) it would be a bit of a pain to sort for it to only still come out kind of wrong.

I can still obviously do it (and probably handle the numeric/"N/A" cases too since I'd have to implement the comparison myself), just questioning if it's worth the effort.

omertuc: For now let's leave it as is, maybe we'll improve it later. I think maybe even compact "0","1","2","3" into "0-3" since the most common case is if one cert was modified, it's likely all other certs in the bundle have also been modified. I don't want our annotation to look too ugly and take up screen space (like managed-fields usually does)

PR #17, opened 2023-09-13
Add ouger as a git submodule

mresvanis:
This PR adds [ouger](https://github.com/rh-ecosystem-edge/ouger) as a git submodule and an additional Dockerfile stage for its building and copying into the recert release image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
-FROM rust:1 AS chef 
-# We only pay the installation cost once, 
+FROM rust:1 AS chef
+# We only pay the installation cost once,
 # it will be cached from the second build onwards
-RUN cargo install cargo-chef 
+RUN cargo install cargo-chef
 WORKDIR app
 
 FROM chef AS planner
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -18,11 +18,17 @@ RUN cargo chef cook --release --recipe-path recipe.json
 COPY . .
 RUN cargo build --release --bin recert
 
+FROM golang:1.19-bookworm as ouger-builder
+COPY . .
```
omertuc, line 22: Avoid wildcard copies because they invalidate the cache on every tiny thing

Dockerfile
```diff
@@ -18,11 +18,17 @@ RUN cargo chef cook --release --recipe-path recipe.json
...
+RUN git submodule update --init --recursive
```
omertuc, line 23: I think it's best to put this responsibility on whoever is running the Dockerfile

omertuc (changes requested):
Thank you, a few caching concerns

omertuc:
Finally https://asciinema.org/a/i8KjIinSiaK8HlBoPmpcK1Cza

mresvanis:
> Finally https://asciinema.org/a/i8KjIinSiaK8HlBoPmpcK1Cza

I'll open a PR with the updated `.gitignore` :+1:

omertuc:
No need, there's already an ignore, I was just happy to delete it from the recert root :)

PR #27, opened 2023-10-02
Adjust cluster-rename arg parsing and error msg

mresvanis:


src/ocp_postprocess/cluster_domain_rename/params.rs
```diff
@@ -8,11 +8,11 @@ pub(crate) struct ClusterRenameParameters {
 
 impl ClusterRenameParameters {
     pub(crate) fn cli_parse(value: &str) -> Result<Self> {
-        let parts = if value.contains(',') { value.split(',') } else { value.split(':') }.collect::<Vec<_>>();
+        let parts = { value.split(',') }.collect::<Vec<_>>();
```
omertuc, line 11: `,` split is the old deprecated behavior, let's remove it and keep just `:`

src/ocp_postprocess/cluster_domain_rename/params.rs
```diff
@@ -8,11 +8,11 @@ pub(crate) struct ClusterRenameParameters {
 
 impl ClusterRenameParameters {
     pub(crate) fn cli_parse(value: &str) -> Result<Self> {
-        let parts = if value.contains(',') { value.split(',') } else { value.split(':') }.collect::<Vec<_>>();
+        let parts = { value.split(':') }.collect::<Vec<_>>();
```
omertuc, line 11:
```suggestion
        let parts = value.split(':').collect::<Vec<_>>();
```

PR #63, opened 2023-11-27
Support symmetric keys

omertuc:
Solves [MGMT-15682](https://issues.redhat.com//browse/MGMT-15682)


omertuc:
We're not gonna implement it for now, as in our main use cases, there are no symmetric keys. We might revisit this in the future

PR #64, opened 2023-11-29
OCPEDGE-654 Fix dep annotations

omertuc:
This should avoid some rollouts

image


romfreiman:
@omertuc can u add OCPEDGE-654 to the title?

PR #66, opened 2023-12-05
Avoid pem::encode_many_config because of duplicate newlines

omertuc:
We re-implement encode_many_config ourselves without the excess newlines


omertuc:
Closing in favor of #67

PR #69, opened 2023-12-08
Sync webhook-authenticator static secret files

omertuc:
The webhook authenticator secret has a kubeConfig field that is too
complicated to handle in recert. We could simply delete it and it will
be reconciled, but that's a bit too slow for us as it causes a
kube-apiserver rollout. To speed things up, we'll just "reconcile" it
ourselves by copying the kubeConfig contents from the kubeConfig file on
disk that we already processed with recert.

Also enable the preserve_order feature of serde_json to make sure field
order is maintained, because those differences also cause rollouts

________________

Older revisions are deleted from etcd, only the latest revision is synced:

image


mresvanis:
> /test baremetalds-sno-recert-cluster-rename

@omertuc I think this test is going to fail until [this](https://github.com/openshift-metal3/dev-scripts/pull/1617) PR is merged.

PR #82, opened 2024-01-26
Implement ocp postprocess hostname rename

mresvanis:
This PR adds the changes required to the `ocp_postprocess` module, in order to skip the unnecessary and time-consuming OCP control-plane component rollouts. The latter get triggered by the node's `hostname` change, which occurs when applying an IBU seed image onto the target SNO.

The changes are:

- [x] add a new recert configuration argument of type `string`, i.e. `--hostname` with which the new `hostname` is provided
    - [ ] we might need to also replace the `NODE_IP` in the etcd related `configmaps`, in addition to the `hostname`, which is currently done in the filesystem by the LCA post-pivot. It's probably a good idea to do it here, so that we can change it both in etcd and the filesystem. In that sense, this argument could be refactored to `--hostname-rename <hostname>:<IP>` (?)
- [x] add a `hostname_rename` mod, which implements the following changes:
    - [x] replace all occurrences of the previous (seed node) `hostname` in `etcd`
        - [x] `secrets/openshift-etcd/etcd-all-certs` and its revisions
        - [x] `operator.openshift.io/kubeapiservers/cluster`
        - [x]  `operator.openshift.io/kubecontrollermanagers/cluster`
        - [x]  `operator.openshift.io/kubeschedulers/cluster`
        - [x] `secrets/openshift-etcd/etcd-peer` and its revisions
        - [x] `secrets/openshift-etcd/etcd-serving`
        - [x] `secrets/openshift-etcd/etcd-serving-metrics`
        - [x] `configmaps/openshift-etcd/etcd-pod` and its revisions
        - [x] `configmaps/openshift-etcd/restore-etcd-pod`
        - [x] `configmaps/openshift-etcd/etcd-scripts`
    - [x] replace all occurrences of the previous (seed node) `hostname` in the filesystem
        - [x] `/etc/kubernetes/static-pod-resources/etcd-*` env var names and values of env vars, i.e. `s/NODE_<hostname>/NODE_<new-hostname>/g; s/<hostname/<new-hostname>/g`
        - [x] `/etc/kubernetes/manifests/etcd-*` env var names and env var values, i.e. `s/NODE_<hostname>/NODE_<new-hostname>/g; s/<hostname>/<new-hostname>/g`
        - [x] `/etc/kubernetes/static-pod-resources/kube-apiserver-pod-*`, i.e. `s/--node-name=<hostname>/--node-name=<new-hostname>/g`
        - [x] `/etc/kubernetes/static-pod-resources/etcd-certs/secrets/etcd-all-certs/*.{crt,key}`
        - [x] `/etc/kubernetes/static-pod-resources/etcd-member/etcd-all-certs/*.{crt,key}`
        - [x] `/etc/kubernetes/static-pod-resources/etcd-pod-*/secrets/etcd-all-certs/*.{crt,key}`
- [ ] replace `hostname` in `etcd` keys and values, that are irrelevant to skipping the control-plane component rollouts but still useful in order to not leave any leftovers when moving to the seed image to the target SNO

### DIFF
--- a/src/ocp_postprocess/hostname_rename/filesystem_rename.rs
+++ b/src/ocp_postprocess/hostname_rename/filesystem_rename.rs
@@ -1,0 +1,260 @@
+use crate::{
+    file_utils::{self, commit_file, read_file_to_string, DRY_RUN},
+    ocp_postprocess::cluster_domain_rename::rename_utils,
+};
+use anyhow::{self, Context, Result};
+use futures_util::future::join_all;
+use serde_json::Value;
+use std::{fs, path::Path, sync::atomic::Ordering::Relaxed};
+
+pub(crate) async fn fix_filesystem_etcd_static_pods(original_hostname: &str, hostname: &str, dir: &Path) -> Result<()> {
+    join_all(file_utils::globvec(dir, "**/etcd-pod.yaml")?.into_iter().map(|file_path| {
+        let etcd_pod_path = file_path.clone();
+        let original_hostname = original_hostname.to_string();
+        let hostname = hostname.to_string();
+        tokio::spawn(async move {
+            async move {
+                let contents = read_file_to_string(&file_path).await.context("reading etcd-pod.yaml")?;
+
+                let mut pod: Value = serde_json::from_str(&contents).context("parsing etcd.yaml")?;
+
+                rename_utils::fix_etcd_static_pod(&mut pod, &original_hostname, &hostname).context("fixing etcd-pod.yaml")?;
+
+                commit_file(file_path, serde_json::to_string(&pod).context("serializing etcd-pod.yaml")?)
+                    .await
+                    .context("writing etcd-pod.yaml to disk")?;
+
+                anyhow::Ok(())
+            }
+            .await
+            .context(format!("fixing etcd-pod.yaml {:?}", etcd_pod_path))
+        })
+    }))
+    .await
+    .into_iter()
+    .collect::<Result<Vec<_>, _>>()?
+    .into_iter()
+    .collect::<Result<Vec<_>>>()?;
+
+    Ok(())
+}
+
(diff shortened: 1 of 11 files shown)


src/ocp_postprocess/hostname_rename/filesystem_rename.rs
```diff
@@ -228,3 +229,35 @@ pub(crate) async fn fix_filesystem_kapi_startup_monitor_configmap_pod_yaml(
...
+                let original_path = file_path.clone();
+                let new_path = file_path
+                    .clone()
+                    .into_os_string()
+                    .into_string()
+                    .unwrap()
+                    .replace(original_hostname, hostname);
+                tokio::spawn(async move {
+                    async move {
+                        fs::rename(original_path, new_path)?;
```
omertuc, line 247: we should have a function similar to commit_file so it can be avoided during a dry run

mresvanis: Of course, I completely neglected the dry run option. I'll try to update it accordingly.

mresvanis: I didn't add a function like `commit_rename` because we only need it here, but I can do that if you think it makes sense for the long term. WDYT?

omertuc: I prefer a separate function just because I don't want to concern rename code with the concept of dry run

omertuc: But I'll just do it separately later as part of a refactor of all this code, too much repetition. I'm afraid to touch it because we have no IBI/IBU tests

mresvanis: Got it, this makes sense, I'm working on adding the recert hostname change 0-rollout CI job (not IBI/IBU at this point, just recert, but I think until we have an IBI/IBU e2e job, we can already increase our confidence starting with this recert specific job).

PR #84, opened 2024-01-31
Report separate running time for RSA key generation and scanning

omertuc:
Even though they happen in parallel, it's interesting to know how long
each of them took separately


omertuc:
e2e-aws-ovn-single-node-recert-serial  OK

PR #85, opened 2024-01-31
Improve error handling

omertuc:
* Ensure openssl version is major 3 (major 1 behaves differently, e.g.
  PKCS#1 RSA private keys by default instead of PKCS#8)

* A few missing contexts


omertuc:
e2e-aws-ovn-single-node-recert-serial  OK

PR #91, opened 2024-02-06
Fix etcd rollout

omertuc:
Typo in hostname rename


mresvanis:
Oh my ... Many thanks @omertuc :pray:

PR #93, opened 2024-02-06
MGMT-16780: rename node name in OCP hostname post-process

mresvanis:
This change adds a step to the OCP hostname post-process. Specifically, it renames the node object and sets its hostname address to the hostname defined by the --hostname arg.

Related LCA PR: https://github.com/openshift-kni/lifecycle-agent/pull/290

### DIFF
--- a/src/etcd_encoding.rs
+++ b/src/etcd_encoding.rs
@@ -4,7 +4,7 @@ use super::protobuf_gen::{
         api::{
             admissionregistration::v1::{MutatingWebhookConfiguration, ValidatingWebhookConfiguration},
             apps::v1::{DaemonSet, Deployment},
-            core::v1::{ConfigMap, Secret},
+            core::v1::{ConfigMap, Node, Secret},
         },
         apimachinery::pkg::runtime::{TypeMeta, Unknown},
     },
@@ -51,6 +51,7 @@ k8s_type!(RouteWithMeta, Route);
 k8s_type!(DaemonsSetWithMeta, DaemonSet);
 k8s_type!(DeploymentWithMeta, Deployment);
 k8s_type!(ConfigMapWithMeta, ConfigMap);
+k8s_type!(NodeWithMeta, Node);
 k8s_type!(SecretWithMeta, Secret);
 k8s_type!(ValidatingWebhookConfigurationWithMeta, ValidatingWebhookConfiguration);
 k8s_type!(MutatingWebhookConfigurationWithMeta, MutatingWebhookConfiguration);
@@ -69,6 +70,7 @@ pub(crate) async fn decode(data: &[u8]) -> Result<Vec<u8>> {
         "Deployment" => serde_json::to_vec(&DeploymentWithMeta::try_from(unknown)?)?,
         "DaemonSet" => serde_json::to_vec(&DaemonsSetWithMeta::try_from(unknown)?)?,
         "ConfigMap" => serde_json::to_vec(&ConfigMapWithMeta::try_from(unknown)?)?,
+        "Node" => serde_json::to_vec(&NodeWithMeta::try_from(unknown)?)?,
         "Secret" => serde_json::to_vec(&SecretWithMeta::try_from(unknown)?)?,
         "ValidatingWebhookConfiguration" => serde_json::to_vec(&ValidatingWebhookConfigurationWithMeta::try_from(unknown)?)?,
         "MutatingWebhookConfiguration" => serde_json::to_vec(&MutatingWebhookConfigurationWithMeta::try_from(unknown)?)?,
@@ -90,6 +92,7 @@ pub(crate) async fn encode(data: &[u8]) -> Result<Vec<u8>> {
     result.extend(
         match kind {
             "ConfigMap" => Unknown::from(serde_json::from_slice::<ConfigMapWithMeta>(data)?),
+            "Node" => Unknown::from(serde_json::from_slice::<NodeWithMeta>(data)?),
             "Route" => Unknown::from(serde_json::from_slice::<RouteWithMeta>(data)?),
             "Secret" => Unknown::from(serde_json::from_slice::<SecretWithMeta>(data)?),
             "Deployment" => Unknown::from(serde_json::from_slice::<DeploymentWithMeta>(data)?),
--- a/src/ocp_postprocess/hostname_rename/etcd_rename.rs
+++ b/src/ocp_postprocess/hostname_rename/etcd_rename.rs
@@ -402,6 +402,28 @@ pub(crate) async fn fix_node_name(etcd_client: &Arc<InMemoryK8sEtcd>, original_h
 
         metadata.insert("name".to_string(), Value::String(hostname.to_string()));
 
+        let annotations = &mut node
+            .pointer_mut("/metadata/annotations")
+            .context("no annotations")?
+            .as_object_mut()
+            .context("/metadata/annotations not a object")?;
+
+        let annotations_to_delete = [
+            "k8s.ovn.org/zone-name",
+            "k8s.ovn.org/remote-zone-migrated",
+            "k8s.ovn.org/host-cidrs",
+            "k8s.ovn.org/l3-gateway-config",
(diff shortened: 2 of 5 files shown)


src/etcd_encoding.rs
```diff
@@ -4,7 +4,7 @@ use super::protobuf_gen::{
         api::{
             admissionregistration::v1::{MutatingWebhookConfiguration, ValidatingWebhookConfiguration},
             apps::v1::{DaemonSet, Deployment},
-            core::v1::{ConfigMap, Secret},
+            core::v1::{ConfigMap, Node, Secret},
```
tsorya, line 7: should not we change ip in node object too?

omertuc: I've been looking into it, looks like we can safely delete these annotations (which contain the IP):

```
k8s.ovn.org/host-cidrs
k8s.ovn.org/l3-gateway-config
k8s.ovn.org/node-chassis-id
k8s.ovn.org/node-mgmt-port-mac-address
k8s.ovn.org/node-primary-ifaddr
```

And they get recreated.

Same thing with the node status - if you delete it it gets recreated immediately

The only trouble I've had was with these:

```
machineconfiguration.openshift.io/currentConfig: rendered-master-482484efc39298d216b84f84e49ce1fb
machineconfiguration.openshift.io/desiredConfig: rendered-master-482484efc39298d216b84f84e49ce1fb
machineconfiguration.openshift.io/desiredDrain: uncordon-rendered-master-482484efc39298d216b84f84e49ce1fb
machineconfiguration.openshift.io/lastAppliedDrain: uncordon-rendered-master-482484efc39298d216b84f84e49ce1fb
machineconfiguration.openshift.io/lastSyncedControllerConfigResourceVersion: "278801"
```

Which contain the old MC rendered hashes. If you delete them, they don't get recreated, MCO doesn't if we do this. So we should leave them as is and just see what happens

mresvanis: Both spot on comments, I was thinking of testing these first and then see what exactly needs to be done. I'll keep you posted, please let me know if you manage to verify any of these.

omertuc:
We also need to delete this line:

https://github.com/rh-ecosystem-edge/recert/blob/88cb8c4811451b2e4b420dd62812bbe0e3c69701/src/ocp_postprocess.rs#L399

src/ocp_postprocess/hostname_rename/etcd_rename.rs
```diff
@@ -387,6 +387,43 @@ pub(crate) async fn fix_etcds_cluster(etcd_client: &Arc<InMemoryK8sEtcd>, hostna
...
+        .as_array_mut()
+        .context("/status/addresses not an array")?;
+
+    addresses
+        .iter_mut()
+        .filter(|address| address["type"] == "Hostname")
+        .for_each(|address: &mut Value| {
+            address
+                .as_object_mut()
+                .unwrap()
```
omertuc, line 416: `unwrap()` on an `Option` that has the potential to be `None` can cause panics, and recert generally never panics, so this should be avoided.

Instead use `anyhow`'s `.context("something")` to convert the `Option<>` type into a `Result<>` (if the `Option<>` is `None` the resulting `Result` will be an `Err`, otherwise if the `Option` is a `Some(_)` the resulting `Result` will be an `Ok(_)` containing the same value as the original `Option`. After this you can `?` to return an error on `Err` from the `for_each`'s lambda, which you should change to a `try_for_each` instead, which would return the first error upon the first error, and wrap the entire `address. ... . insert` expression in an `Ok`, then you can `?` on the `try_for_each`'s return value

mresvanis: Awesome, I'll try to incorporate everything you just explained :pray:

mresvanis: Updated PTAL (although we probably won't keep this code, as the status should be updated by the respective controller without any issues)

omertuc: Ah lol didn't notice it's part of the status

mresvanis: :D absolutely no worries TIL :muscle:

src/ocp_postprocess/hostname_rename/etcd_rename.rs
```diff
@@ -387,6 +387,32 @@ pub(crate) async fn fix_etcds_cluster(etcd_client: &Arc<InMemoryK8sEtcd>, hostna
     Ok(())
 }
 
+pub(crate) async fn fix_node_name(etcd_client: &Arc<InMemoryK8sEtcd>, original_hostname: &str, hostname: &str) -> Result<()> {
+    let mut k8s_resource_location = K8sResourceLocation::new(None, "Minion", original_hostname, "v1");
```
omertuc, line 391: Let's change the [plural](https://github.com/rh-ecosystem-edge/recert/blob/dc0186f23a561ba65e1a399a9a383775d9840cb0/src/cluster_crypto/locations.rs#L54) to minions so we can do this:

```suggestion
    let mut k8s_resource_location = K8sResourceLocation::new(None, "Node", original_hostname, "v1");
```

src/ocp_postprocess/hostname_rename/etcd_rename.rs
```diff
@@ -387,6 +387,32 @@ pub(crate) async fn fix_etcds_cluster(etcd_client: &Arc<InMemoryK8sEtcd>, hostna
...
+        .context("no metadata")?
+        .as_object_mut()
+        .context("/metadata not an object")?;
+
+    metadata.insert("name".to_string(), Value::String(hostname.to_string()));
+
+    k8s_resource_location.name = hostname.to_string();
+    put_etcd_yaml(etcd_client, &k8s_resource_location, node).await?;
+
+    let previous_key = format!("/kubernetes.io/minions/{original_hostname}");
```
omertuc, line 407:
```suggestion
    let previous_key = format!("/kubernetes.io/minions/{original_hostname}").as_etcd_key();

    k8s_resource_location.name = hostname.to_string();
    put_etcd_yaml(etcd_client, &k8s_resource_location, node).await?;
```

mresvanis: Niiice, thanks, that's a lot better :)

mresvanis:
@omertuc @tsorya I have tested these changes alongside https://github.com/openshift-kni/lifecycle-agent/pull/290, networking works as expected and there are no control-plane rollouts. PTAL

mresvanis:
/hold 

Until we ensure on how to sync with [this](https://github.com/openshift-kni/lifecycle-agent/pull/290) PR, as if we don't leave the seed node object in the seed image, recert will currently fail.

mresvanis:
/unhold

as we now [ensure](https://github.com/openshift-kni/lifecycle-agent/pull/290) that we won't break the IBU flow until the LCA seed generation doesn't delete the seed node object - [PR](https://github.com/openshift-kni/lifecycle-agent/pull/290)

omertuc:
merge?

mresvanis:
> merge?

Not yet, sorry for the delay. I run another IBU flow and it seems that we need to do more steps to have a healthy cluster after renaming the node object...

mresvanis:
/hold

Until we ensure that networking will successfully reconcile after renaming the node object.

mresvanis:
The respective recert CI jobs won't help us here, as we currently [still delete the node object before running recert](https://github.com/openshift/release/blob/master/ci-operator/step-registry/baremetalds/sno/rename/baremetalds-sno-rename-commands.sh#L170). After merging this PR we should update that CI job to reflect the new flow, i.e. of NOT deleting the node object, but letting recert rename it.

mresvanis:
See this PR's [description](https://github.com/openshift-kni/lifecycle-agent/pull/290#issue-2121798613) for more details.

### TL;DR

Recert already deletes the node object in its OCP post-process flow and that's what's needed to skip the etcd rollout caused by the empty `etcd-all-certs` secret in the seed image.

PR #99, opened 2024-02-14
MGMT-16875: Skip kube-apiserver rollout in OCP 4.15

mresvanis:
This change skips a `kube-apiserver` rollout in OCP 4.15, which is triggered because of one of the kube-apiserver config arguments.  

Specifically,  the `service-account-jwks-uri` config arg contains a URI with the `api` label, instead of the `api-int` one and the cluster domain. To ensure compatibility with previous (and potentially new) OCP versions, we now match both the `api` and `api-int` labels when replacing the cluster domain in that field.


mresvanis:
I have tested this change in the IBU flow successfully with an LCA seed image based on OCP `4.14.6` and `4.15.0-rc.5`, i.e. no kube-apiserver rollout in either OCP version.

PR #105, opened 2024-02-22
MGMT-17018: Fix oauth-apiserver deployment IP replacement

mresvanis:
When trying to find the --etcd-servers arg of the oauth-apiserver deployment container command, in order to replace the seed IP with the recipient IP, recert did not include single quotes in the find string. The oauth-server regexp-s for shell escapes and encodes the arg value that matches. That made IPv6 IPs, which contain '[' (and ']'), be missed by recert previously.

- https://github.com/openshift/cluster-authentication-operator/blob/master/pkg/controllers/common/arguments/arguments.go#L13


omertuc:
Merge whenever you want

PR #108, opened 2024-02-27
MGMT-17058: Fix IPv6 replacement by adding brackets when necessary 

mresvanis:
This PR fixes the IP replacement when an IPv6 is provided or getting replaced. IPv6 IPs should be enclosed in `[ ] ` when used in URLs, but OpenShift in some cases uses brackets for IPv6 IPs also when found in non-URL fields. For example, in the env vars of the etcd container of the `etcd-pod` manifest:

```YAML
- name: "NODE_sno_test_redhat_com_IP"
  value: "[2710:20:0:18::8]"
```

This change ensures that we respect the format of the fields in OCP containing IPv6 IPs (with or without `[ ]`). It also ensures that we support the following cases properly when replacing the  IP:

- seed IPv4 -> recipient IPv4
- seed IPv4 -> recipient IPv6
- seed IPv6 -> recipient IPv4
- seed IPv6 -> recipient IPv6

### DIFF
--- a/src/ocp_postprocess/cluster_domain_rename/rename_utils.rs
+++ b/src/ocp_postprocess/cluster_domain_rename/rename_utils.rs
@@ -110,6 +110,14 @@ pub(crate) fn fix_api_server_arguments_ip(config: &mut Value, original_ip: &str,
         .context("etcd-servers not an array")?
         .clone();
 
+    let original_ip = if original_ip.contains(':') {
+        format!("[{original_ip}]")
+    } else {
+        original_ip.to_string()
+    };
+
+    let ip = if ip.contains(':') { format!("[{ip}]") } else { ip.to_string() };
+
     let new_etcd_servers = original_etcd_servers
         .iter()
         .map(|etcd_server| {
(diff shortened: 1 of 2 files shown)


src/ocp_postprocess.rs
```diff
@@ -61,7 +62,16 @@ pub(crate) async fn ocp_postprocess(
     }
 
     if let Some(ip) = ip {
-        ip_rename(in_memory_etcd_client, ip, static_dirs, static_files)
+        // OpenShift uses `[ ]` to enclose IPv6 IPs also in non-URL fields. Let us add the brackets
```
omertuc, line 65: Is it truly correct for all places where we replace the IP?

mresvanis: That's a great question, it seems to be that way, I'll put a list here so that you can review the location as well.

mresvanis: This is what we have in etcd for an IPv6 IP with brackets:

```shell
/kubernetes.io/configmaps/openshift-apiserver/config
/kubernetes.io/configmaps/openshift-kube-apiserver/config-7
/kubernetes.io/configmaps/openshift-kube-apiserver/config-9
/kubernetes.io/configmaps/openshift-kube-apiserver/config-6
/kubernetes.io/configmaps/openshift-kube-apiserver/config
/kubernetes.io/configmaps/openshift-kube-apiserver/config-10
/kubernetes.io/configmaps/openshift-kube-apiserver/config-8
/kubernetes.io/configmaps/openshift-etcd/etcd-pod-3
/kubernetes.io/configmaps/openshift-etcd/etcd-scripts
/kubernetes.io/configmaps/openshift-etcd/etcd-pod-2
/kubernetes.io/configmaps/openshift-etcd/etcd-pod
/kubernetes.io/configmaps/openshift-etcd/restore-etcd-pod
/kubernetes.io/deployments/openshift-oauth-apiserver/apiserver
/kubernetes.io/operator.openshift.io/openshiftapiservers/cluster
/kubernetes.io/operator.openshift.io/authentications/cluster
/kubernetes.io/operator.openshift.io/kubeapiservers/cluster
/kubernetes.io/pods/openshift-kube-controller-manager/kube-controller-manager-another-hostname
/kubernetes.io/pods/openshift-oauth-apiserver/apiserver-6889868855-nhkqz
/kubernetes.io/pods/openshift-etcd/etcd-another-hostname
/kubernetes.io/replicasets/openshift-oauth-apiserver/apiserver-5c6d74c9db
/kubernetes.io/replicasets/openshift-oauth-apiserver/apiserver-6889868855
```

I believe we cover only those in recert currently. Wherever else I found the IP without brackets is not in static pods or in resources in etcd we should care (famous last words :) ).

PR #110, opened 2024-02-28
Change a cluster's trust bundle

omertuc:
# Main change

Support changing a cluster's additional trust bundle. Changes all locations where the additional trust bundle is stored in.

If an existing trust bundle is not found, this will cause an error, as creating the relevant resources is beyond the scope of this tool.

The trust bundle's validity will not be checked. When using a RECERT_CONFIG file, raw PEMS can be used instead of a path to a trust bundle file.

When using this feature it is recommended to also run the `update-ca-trust` script after running recert to ensure that the trust bundle is properly updated in all locations.

# Other changes

* Created `./hack/` directory to store some certs used during `./run_seed.sh`

* Deprecated --static-files and --static-dirs, which were used for both recert and rename. Now `--crypto-dir` and `--crypto-file` will be used for recert while `--cluster-customization-dir` and `--cluster-customization-file` will be used for rename (aka cluster customization). This was needed because /etc/pki is full of certs we discover and fail to process during recert, but we do need to process /etc/pki for editing the additional trust bundle cluster customization. Using `--additional-trust-bundle` along with `--static-*` will cause an error. The old behavior for `--static-files` and `--static-dirs` is maintained for backwards compatibility, but they cannot be used along with the new flags.

* Made ConfigPath a less leaky abstraction for ClioPath and moved its relevant code to its own module `path` (under `config`)

* Renamed many `cli_parse` functions to `parse` as those functions were used outside of CLI parsing as well (during config file parsing)

* Refactored config parsing into topical functions because that functions was getting a bit too long


src/ocp_postprocess/additional_trust_bundle/filesystem_rename.rs
```diff
@@ -0,0 +1,95 @@
...
+                tokio::spawn(async move {
+                    async move {
+                        commit_file(file_path, additional_trust_bundle.clone())
+                            .await
+                            .context("writing to disk")?;
+
+                        anyhow::Ok(())
+                    }
+                    .await
+                    .context(format!("fixing etcd-pod.yaml {:?}", crt_file_path))
```
mresvanis, line 25:
```suggestion
                    .context(format!("fixing CA bundle certificate {:?}", crt_file_path))
```

src/ocp_postprocess/additional_trust_bundle.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    Ok(())
+}
+
+async fn fix_dir_resources(additional_trust_bundle: &str, new_merged_bundle: &str, dir: &Path) -> Result<()> {
+    // NOTE: This only fixes the trust anchors, the user should run "update-ca-trust" to fully
+    // update the system trust store after this change (this is also what MCO does).
+    filesystem_rename::fix_filesystem_ca_trust_anchors(additional_trust_bundle, dir)
+        .await
+        .context("fixing etcd static pods")?;
```
mresvanis, line 47:
```suggestion
        .context("fixing CA trust bundle anchors")?;
```

mresvanis:
/lgtm

I tested it with LCA `release-4.15`, in order to verify backwards compatibility and everything looks good in that respect as well :+1:

PR #112, opened 2024-02-29
MGMT-17057: Update auth annotations when regenerating certs found in etcd

mresvanis:
This PR ensures that when recert regenerates certs that are found in etcd and include the `auth.openshift.io/certificate-not{after,before}` annotations, it also updates the latter according to their respective validity period.

### DIFF
--- a/src/file_utils.rs
+++ b/src/file_utils.rs
@@ -1,9 +1,11 @@
 use crate::cluster_crypto::{
+    certificate::Certificate,
     locations::{FileLocation, JsonLocation, LocationValueType},
     pem_utils,
 };
 use anyhow::{bail, Context, Result};
 use base64::{engine::general_purpose::STANDARD as base64_standard, Engine as _};
+use chrono::{DateTime, SecondsFormat, Utc};
 use serde_json::Value;
 use std::{
     path::{Path, PathBuf},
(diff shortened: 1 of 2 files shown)


src/file_utils.rs
```diff
@@ -235,3 +237,51 @@ pub(crate) fn add_recert_edited_annotation(_resource: &mut Value, _yaml_location
...
+        String::from("auth.openshift.io/certificate-not-after"),
+        Value::String(time_rfc3339(&certificate.tbs_certificate.validity.not_after)),
+    );
+    annotations.insert(
+        String::from("auth.openshift.io/certificate-not-before"),
+        Value::String(time_rfc3339(&certificate.tbs_certificate.validity.not_before)),
+    );
+
+    Ok(())
+}
```
omertuc, line 287: A bit simplified implementation (untested)

```suggestion
pub(crate) fn update_auth_certificate_annotations(resource: &mut Value, certificate: &Certificate) -> Result<()> {
    if let Some(annotations) = resource.pointer_mut("/metadata/annotations") {
        let annotations = annotations.as_object_mut().context("annotations must be an object")?;

        let cert: &x509_certificate::X509Certificate = &certificate.cert;
        let certificate: &x509_certificate::rfc5280::Certificate = cert.as_ref();

        if let Some(not_before) = annotations.get_mut("auth.openshift.io/certificate-not-before") {
            *not_before = Value::String(time_rfc3339(&certificate.tbs_certificate.validity.not_before));
        }

        if let Some(not_after) = annotations.get_mut("auth.openshift.io/certificate-not-after") {
            *not_after = Value::String(time_rfc3339(&certificate.tbs_certificate.validity.not_after));
        }
    }

    Ok(())
}
```

mresvanis: Niiiiiice, testing it now

mresvanis: I think it's not exactly the same (the previous implementation was only changing the annotations when both {after,before} were present), but it works perfectly, it's idiomatic and awesomely more readable. So let's go with that :+1:

PR #120, opened 2024-03-15
Run cargo test as part of the CI

omertuc:
I've added some unit tests recently, but they were not run as part of our GitHub actions. This commit that.

Also fix some broken unit tests. The unit tests were correct but sometimes they would race with each other and fail. I added the `serial_test` crate so we can annotate the tests with `#[serial]` to ensure they don't run together (as these tests modify a global variable).

Also fixed some unused var warning that seems to have only started showing up in recent versions of rust/clippy.


mresvanis:
I'll add the unit test execution in Prow as well :+1:

PR #123, opened 2024-03-28
MGMT-17319: Fix etcd secrets annotations for OCP >4.16

mresvanis:
In OCP 4.16 the cluster-etcd-operator moved to using [library-go/certrotation](https://github.com/openshift/library-go/tree/master/pkg/operator/certrotation) and that leads to the following etcd TLS secrets having annotations that include the IP and the hostname:

- `openshift-etcd/etcd-peer-<hostname>`
- `openshift-etcd/etcd-serving-<hostname>`
- `openshift-etcd/etcd-serving-metrics-<hostname>`

The respective annotations are:
- `auth.openshift.io/certificate-hostnames` - this includes the IP
- `openshift.io/description` - this includes the hostname

Recert now replaces the IP and the hostname in those annotations, in order to skip an additional etcd rollout that is triggered because of the former.

Resources:
- https://github.com/openshift/cluster-etcd-operator/pull/1194
- https://github.com/openshift/library-go/blob/f840ed4b8db36652d101604930ac32df568e583d/pkg/operator/certrotation/target.go#L311
- https://github.com/openshift/cluster-etcd-operator/blob/eeef8033eb058b0f9899c7b16aead237d5b84462/pkg/tlshelpers/tlshelpers.go#L237-L240
    - https://github.com/openshift/library-go/blob/f840ed4b8db36652d101604930ac32df568e583d/pkg/operator/certrotation/annotations.go#L31-L34  
    - https://github.com/openshift/api/blob/master/annotations/annotations.go#L24

### DIFF
--- a/src/ocp_postprocess.rs
+++ b/src/ocp_postprocess.rs
@@ -76,14 +76,14 @@ async fn run_cluster_customizations(
             .context("renaming cluster")?;
     }
 
+    if let Some(ip) = &cluster_customizations.ip {
+        ip_rename(in_memory_etcd_client, ip, dirs, files).await.context("renaming IP")?;
+    }
+
     if let Some(hostname) = &cluster_customizations.hostname {
         hostname_rename(in_memory_etcd_client, hostname, dirs, files)
             .await
             .context("renaming hostname")?;
-    }
-
-    if let Some(ip) = &cluster_customizations.ip {
-        ip_rename(in_memory_etcd_client, ip, dirs, files).await.context("renaming IP")?;
     }
 
     if let Some(kubeadmin_password_hash) = &cluster_customizations.kubeadmin_password_hash {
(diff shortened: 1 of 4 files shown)


mresvanis:
/hold

I just manually tested this change and it seems there is still something missing in order to skip this rollout. I'm on it and I'll update this PR accordingly.

mresvanis:
@omertuc PTAL I've updated the PR only to correctly handle the type [here](https://github.com/rh-ecosystem-edge/recert/pull/123/files#diff-721e56c7bf66143c3ca439be811b6c7beea5b83d7f4e5b02f4b83b19146f4058R215) and [here](https://github.com/rh-ecosystem-edge/recert/pull/123/files#diff-94a20adfc7bc6fad4b5bf9fdc09b76021b76a4de6179b1225e53b5311444d89eR306).

PR #125, opened 2024-04-05
Include error in recert summary

omertuc:
When recert runs into an error, it returns it as a `Result` from the
`main` function. This causes the Rust runtime to print the error trace.

However, the error does not get included in the recert summary files. In
the summary, you only see the logs, which makes it look like no error
occurred.

This commit makes it so that when recert runs into an error the error
will be included in summary files

### DIFF
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -82,12 +82,14 @@ struct Summary {
     recert_config: RecertConfig,
     logs: Vec<String>,
     run_times: Option<RunTimes>,
+    error: Option<String>,
 }
 
 pub(crate) fn generate_summary(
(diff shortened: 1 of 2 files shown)


omertuc:
hmm why didn't the jobs run

mresvanis:
> hmm why didn't the jobs run

https://prow.ci.openshift.org/?job=*pull*recert-main*

mresvanis:
> > hmm why didn't the jobs run
> 
> https://prow.ci.openshift.org/?job=*pull*recert-main*

I apologize, I shouldn't have missed this when reviewing https://github.com/openshift/release/pull/50476 :see_no_evil:

PR #127, opened 2024-04-10
MGMT-16777: Compute the spec-hash annotation for the control-plane components

mresvanis:
This change ensures that rollouts after recert for the following components are skipped:

- [openshift-apiserver/openshift-apiserver](https://github.com/openshift/cluster-openshift-apiserver-operator/blob/release-4.16/bindata/v3.11.0/openshift-apiserver/deploy.yaml)
- [openshift-oauth-apiserver/oauth-apiserver](https://github.com/openshift/cluster-authentication-operator/blob/release-4.16/bindata/oauth-apiserver/deploy.yaml)
- [openshift-authentication/oauth-openshift](https://github.com/openshift/cluster-authentication-operator/blob/release-4.16/bindata/oauth-openshift/deployment.yaml)

The end goal being the reduction of the stabilization time of the Ingress and the Authentication cluster operators.

Specifically, those rollouts are triggered by the difference in annotations on the respective deployments. The latter include a [spec-hash](https://github.com/openshift/library-go/blob/738f3fa4ccf8f4d8c9ff2fd8c31bc68a69f00ab9/pkg/operator/resource/resourceapply/apps.go#L59-L70) (SHA256 of the deployment.spec) annotation and multiple ConfigMap and Secret [fnv32 hash](https://github.com/openshift/library-go/blob/738f3fa4ccf8f4d8c9ff2fd8c31bc68a69f00ab9/pkg/operator/resource/resourcehash/as_configmap.go#L20-L35) annotations of the objects that the cluster-operators are watching.

In order to correctly compute the spec-hash annotations, we need to ensure that whatever the respective cluster-operators compute the hash on, is what we use as well. Because the operators include a mix of YAML and Go code for the specs they compute the spec-hash on (e.g. [this](https://github.com/openshift/cluster-authentication-operator/blob/release-4.16/pkg/operator/workload/sync_openshift_oauth_apiserver.go#L131-L227) is what the cluster-authentication-operator does for the oauth-apiserver), we try to do the same with a hack script written in Go. The latter downloads the required deployment YAML manifests from the specific OCP release branch, encodes them in JSON (mainly because JSON encoding between Go and Rust are quite different), adds basic spec fields that can be hardcoded and leaves some templated spec fields to be filled in in Rust code (e.g. container image pull-specs, revisions, etc.). The fnv32 hash computations are also adjusted to follow the Go JSON encoding practices.


mresvanis:
@omertuc this is still a WIP, as I have to fill in a few gaps, but you can already review IMHO, because it is a big change and also I'm not 100% sure we would like to add this Go "script", maybe you have a better idea :man_shrugging: .

run_seed.sh
omertuc, line 1: committed by mistake?

src/ocp_postprocess.rs
```diff
@@ -56,13 +59,39 @@ pub(crate) async fn ocp_postprocess(
...
     .context("fixing dep annotations for openshift-oauth-apiserver")?;
 
+    fix_deployment_spec_hash_annotation(
+        in_memory_etcd_client,
+        K8sResourceLocation::new(Some("openshift-oauth-apiserver"), "Deployment", "apiserver", "v1"),
+    )
+    .await
+    .context("fixing spec-hash annotation for openshift-oauth-apiserver")?;
+
+    // TODO: fix spec-hash for openshift-authentication/oauth-openshift
```
omertuc, line 83: Please state reason why it's commented out above the comment

src/ocp_postprocess.rs
```diff
@@ -218,6 +247,107 @@ pub(crate) async fn fix_olm_secret_hash_annotation(in_memory_etcd_client: &Arc<I
     Ok(())
 }
 
+fn verbosity(log_level: &str) -> String {
```
omertuc, line 250: Just to emphasize that this is not recert's own log verbosity

```suggestion
fn operator_log_verbosity(log_level: &str) -> String {
```

src/ocp_postprocess.rs
```diff
@@ -218,6 +247,107 @@ pub(crate) async fn fix_olm_secret_hash_annotation(in_memory_etcd_client: &Arc<I
...
+    match log_level {
```
omertuc, line 251: Add comment mentioning this follows https://github.com/openshift/library-go/blob/15d11d2f6bfcbe15679acd184ac69c77aa2e65bc/pkg/operator/loglevel/util.go#L13-L27

src/ocp_postprocess.rs
```diff
@@ -218,6 +247,107 @@ pub(crate) async fn fix_olm_secret_hash_annotation(in_memory_etcd_client: &Arc<I
...
+                .as_object_mut()
+                .context("spec.observedConfig.oauthAPIServer.apiServerArguments not an object")?;
+            args.insert("v".to_string(), serde_json::Value::String(verbosity.clone()));
+
+            fix_openshift_oauth_apiserver_spec_hash_annotation(
+                metadata_annotations,
+                revision,
+                &verbosity,
+                image,
+                encode_with_delimeter(args.clone(), r" \\\n  ").unwrap().as_str(),
```
omertuc, line 339: no unwrapping

src/ocp_postprocess.rs
```diff
@@ -292,14 +422,29 @@ async fn fix_dep_annotations(
...
+            .as_object()
+            .context("data not an object")?
+            .clone();
+
+        let data_json = if resource_k8s_resource_location.kind == "secret" {
+            // https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/encoding/json/encode.go;l=751-753
+            // https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/encoding/json/encode.go;l=790
+            let sorted: BTreeMap<_, _> = data
+                .iter()
+                .map(|(k, v)| (k, base64_standard.encode(serde_json::from_value::<Vec<u8>>(v.clone()).unwrap())))
```
omertuc, line 439: no unwrapping

src/ocp_postprocess.rs
```diff
@@ -310,6 +455,164 @@ async fn fix_dep_annotations(
     Ok(())
 }
 
+// https://github.com/openshift/cluster-authentication-operator/blob/master/pkg/controllers/common/arguments/arguments.go#L49-L62
+fn shell_escape(s: &str) -> String {
+    let re = regex::Regex::new(r"[^\w@%+=:,./-]").unwrap();
```
omertuc, line 460: no unwrapping

src/ocp_postprocess.rs
```diff
@@ -310,6 +455,164 @@ async fn fix_dep_annotations(
...
+    if re.is_match(s) {
+        format!("'{}'", s.replace('\'', "'\"'\"'"))
+    } else {
+        s.to_string()
+    }
+}
+
+fn trim_quotes(s: &str) -> String {
+    s.strip_prefix('"').unwrap().strip_suffix('"').unwrap().to_string()
```
omertuc, line 469: no unwrapping

omertuc (changes requested):
Surface level review

src/ocp_postprocess.rs
mresvanis: That's still WIP, as I have to compute [this](https://github.com/openshift/cluster-authentication-operator/blob/master/pkg/controllers/deployment/default_deployment.go#L54) first. So I will either properly implement it (if I succeed in that computation) or I'll remove this altogether :+1:

src/ocp_postprocess.rs
```diff
@@ -218,6 +236,140 @@ pub(crate) async fn fix_olm_secret_hash_annotation(in_memory_etcd_client: &Arc<I
...
+                .as_array()
+                .context("spec.template.spec.containers[0].env not an array")?;
+
+            let value = env.iter().find_map(|var| {
+                (var.as_object()?.get("name") == Some(&serde_json::Value::String("KUBE_APISERVER_OPERATOR_IMAGE".to_string())))
+                    .then_some(var.get("value")?)
+            });
+
+            ensure!(value.is_some(), "expected KUBE_APISERVER_OPERATOR_IMAGE to be in env vars");
+            let kube_apiserver_operator_image = value.unwrap().as_str().context("value not a string")?;
```
omertuc, line 329:
```suggestion
            let kube_apiserver_operator_image = value.context("expected KUBE_APISERVER_OPERATOR_IMAGE to be in env vars")?.as_str().context("value not a string")?;
```

src/ocp_postprocess.rs
mresvanis: Any ideas on how to get rid of this `unwrap`? I've been unsuccessfully playing around with it for a bit.

omertuc: Return a vector of results from the map, instead of a vector of strings

A vector of `Result<T>`'s knows how to be collected into a `Result<Vec<T>>`


```rust
        let data_json = if resource_k8s_resource_location.kind == "secret" {
            // https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/encoding/json/encode.go;l=751-753
            // https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/encoding/json/encode.go;l=790
            let sorted: BTreeMap<_, _> = data
                .iter()
                .map(|(k, v)| {
                    Ok((
                        k,
                        base64_standard.encode(serde_json::from_value::<Vec<u8>>(v.clone()).context("oops")?),
                    ))
                })
                .collect::<Result<Vec<_>>>()?
                .into_iter()
                .collect();
```

mresvanis: Thank you for the detailed explanation!

mresvanis:
/hold 

When testing it manually there seems to be an issue with the spec-hash again, leading to the 2 components getting new rollouts.

mresvanis:
/unhold

@omertuc I fixed the issue, I fetched by mistake the `spec.operatorLogLevel` instead of the `spec.logLevel` for the `authentication.operator.io/cluster` and the `openshiftapiserver.operator.io/cluster` CRs.

src/ocp_postprocess.rs
```diff
@@ -218,6 +236,156 @@ pub(crate) async fn fix_olm_secret_hash_annotation(in_memory_etcd_client: &Arc<I
...
+
+            let value = env.iter().find_map(|var| {
+                (var.as_object()?.get("name") == Some(&serde_json::Value::String("KUBE_APISERVER_OPERATOR_IMAGE".to_string())))
+                    .then_some(var.get("value")?)
+            });
+
+            let kube_apiserver_operator_image = value
+                .context("expected KUBE_APISERVER_OPERATOR_IMAGE to be in env vars")?
+                .as_str()
+                .context("value not a string")?;
```
omertuc, line 316: Maybe rust_analyzer extract function these (selected) lines into a function called `get_kube_apiserver_operator_image`?

src/ocp_postprocess.rs
```diff
@@ -218,6 +236,156 @@ pub(crate) async fn fix_olm_secret_hash_annotation(in_memory_etcd_client: &Arc<I
...
+            .await?
+            .context("couldn't find openshiftapiserver.operator/cluster resource")?;
+
+            let log_level = operator_log_level(
+                apiserver
+                    .pointer("/spec/logLevel")
+                    .context("no spec.logLevel")?
+                    .as_str()
+                    .context("spec.logLevel")?,
+            );
```
omertuc, line 331: And extract these into `get_openshift_apiserver_operator_loglevel`

omertuc (changes requested):
Looks good, a few nits

src/ocp_postprocess.rs
mresvanis: Absolutely, I'll update this.

src/ocp_postprocess.rs
mresvanis: Totally agree, I'll update this.

hack/assets/sync.go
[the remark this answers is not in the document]
mresvanis: Of course, `README`s are necessary :+1:  

I'm also thinking about adding a CI job (or maybe even a Github action?) to check if the downloaded deployment manifests are up-to-date (depending on the git branch). WDYT?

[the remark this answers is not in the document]
omertuc: Wouldn't the rollout CI catch this anyway?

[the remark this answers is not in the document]
mresvanis: Yes, you're right, that's one sure way to catch this. I'll first add the rollout checks in CI and then we'll see if we ever need anything else :+1:

src/ocp_postprocess.rs
```diff
@@ -218,6 +236,156 @@ pub(crate) async fn fix_olm_secret_hash_annotation(in_memory_etcd_client: &Arc<I
...
+            let deployment = get_etcd_json(etcd_client, location)
+                .await?
+                .context(format!("couldn't find {}", location))?;
+
+            let env = deployment
+                .pointer("/spec/template/spec/containers")
+                .context("no spec.template.spec.containers")?
+                .as_array()
+                .context("spec.template.spec.containers not an array")?
+                .first()
```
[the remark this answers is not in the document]
mresvanis: I thought of using [find](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find) instead.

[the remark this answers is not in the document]
omertuc: That works

omertuc:
unhold if you don't have any more changes you want to do

mresvanis:
/unhold

@omertuc I will indeed make additions towards skipping the oauth-server rollout (which is not included here, but it's relevant to these changes) and supporting envs with proxy enabled for the openshift-apiserver and oauth-apiserver rollouts. 

But I'll open 2 additional PRs for these specifically, we can already merge these changes safely :+1:

omertuc:

omertuc:

PR #128, opened 2024-04-10
Don't use `unwrap`

omertuc:
Enabled `+#![deny(clippy::unwrap_used, clippy::expect_used, clippy::panic)]`

(but also `allow-unwrap-in-tests = true` because it's too strict for tests)

This means that `unwrap`, `expect` and `panic` are not allowed in the project, unless the programmer states `#[allow(clippy::unwrap_used)]`, preferably with a comment explaining why the `unwrap`/`expect` could never panic.

Removed all instances of `unwrap` which had better alternatives, but left some that were reasonable.

This should stop recert from causing panics (unfortunately dependencies might still panic)

Also removed references to `anyhow::Error` from return values, as this is unnecessarily verbose and can be inferred implicitly. It's constantly being added by the rust_analyzer's extract_function code action, and I forget to remove it.

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,0 +1,1 @@
+allow-unwrap-in-tests = true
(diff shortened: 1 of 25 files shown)


mresvanis:

PR #130, opened 2024-04-18
Add support for machine-network-cidr rename

omertuc:
# Background

The `cluster-config-v1` configmaps contain a copy of
`install-config.yaml` that was used for installation. In that
install-config, one of the fields is the machine-network CIDR.

The cluster-network-operator reads the machine-network field from the
`cluster-config-v1` configmaps while it tries to determine the final
value for `noProxy`, so that means we must ensure those configmaps have
the current machine network value.

# Current support

Currently recert supports replacing the entire install-config in those
configmaps. This is good for IBU, where we already have a cluster we can
take the entire install-config from.

# Issue

In IBI, we would have to fake such install-config for the fresh cluster.
This is inconvenient.

# Solution

Add a new `--machine-network-cidr` rename option to recert, which would
replace just the machine-network-cidr in the existing (seed)
install-config, rather than forcing the user to supply the entire
machine-config.


run_seed.sh
```diff
@@ -95,7 +95,42 @@ else
...
+  networkType: OVNKubernetes
+  serviceNetwork:
+  - 172.30.0.0/16
+platform:
+  none: {}
+publish: External
+pullSecret: ""
+sshKey: ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQDThIOETj6iTvbCaNv15tZg121nWLcwtJuZofc1QS5iAdw8C8fN2R39cSW/ambikl2Fr6YNBBVR3znbtmattOyWyxAOFUfdY0aw0MqZb4LWLf04q6X0KsWIYWaV3ol0KxTzgvX38i/IU42XQfJwMYFE8dQ15TZ7l+FTKKi3SUPXLuy/9CXRfaCDZ2dKMcCkelkTr0KR1HdjiKQ86rMfk9JUbAf7D29aAQq4h1WNnHMM9vnbqN7MW9L8ebn/lCTJjGQ56r0UmurgyIEMt0P+CGp1e4AUNKYsPoYFB0GNwUkr/rB8LeuCOaZcoWdYXlUJaN45GjtCDon56+AoMA9V8tYkV6HqyFwGQjoGKI1cRCHXDJnGyAbMd9OK94TWJmNvtdHkbSURHyw2G7otZpAkRuEvMP0C7R+3JmuxrDA8yaUgWvgccqGcmFl1krClksW6KrAXNlwhZ4QOAMhDrXwwPfOOQoG82zPpg+g9gZQIhkro1Cje4bmWz5z5fiuDloTq1vc=
+  root@edge-01.edge.lab.eng.rdu2.redhat.com' \
+		--machine-network-cidr '192.168.127.0/24' \
```
eranco74, line 133: I think the actual case we want to cover is  `--machine-network-cidr` and no `--install-config`

omertuc: I just like that `./run_seed.sh` uses all the options, as a form of example

omertuc:

mresvanis (approved):

omertuc:

omertuc:

PR #131, opened 2024-04-22
OCPBUGS-32514: Fix recert failing because of non-existent managedFields

mresvanis:
This PR makes the `etcd-all-certs` secret data key replacement in the OCP post-process hostname rename flow optional, as there are cases where the `managedFields` are missing. In the latter case, recert would fail with the respective error.


omertuc:

omertuc:

PR #132, opened 2024-04-25
Retry on etcd too many requests error

omertuc:
Solves [MGMT-17651](https://issues.redhat.com//browse/MGMT-17651)

# tl;dr

A fix for a rare error:

```
Error: finalizing

Caused by:
    0: commiting etcd cache to actual etcd
    1: grpc request error: status: Unknown, message: "etcdserver: too many requests", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }
```

# Background

When committing our in-memory etcd representation to actual etcd, we send all delete requests concurrently (we have many).

# Issue

Sometimes this leads to us receiving an error from etcd which says "etcdserver: too many requests". Recert treated this error as a hard error and as a result it exits.

# Solution

Compare the error string to this exact phrasing (as there doesn't seem to be a more robust error code we can check, the code just says `Unknown`), and if we encounter it, just repeat the request again. Eventually hopefully all requests should go through.

### DIFF
--- a/src/k8s_etcd.rs
+++ b/src/k8s_etcd.rs
@@ -59,7 +59,19 @@ impl InMemoryK8sEtcd {
                     let key = key.clone();
                     let etcd_client = Arc::clone(etcd_client);
                     tokio::spawn(async move {
-                        etcd_client.kv_client().delete(key.as_bytes(), None).await?;
+                        loop {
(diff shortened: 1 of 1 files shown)


omertuc:
Checking clean CI in #134

PR #136, opened 2024-05-07
Fix recert pattern for etcd cluster-backup.sh

tsorya:


omertuc:
The CI is currently broken

PR #140, opened 2024-05-10
Separate user-ca-bundle from proxy bundle

omertuc:
# Background / Context

We recently added a feature to change a cluster's trust bundle

# Issue / Requirement / Reason for change

The feature was incomplete, it didn't distinguish between user-ca-bundle and proxy bundle, both of which could potentially be separate.

In overview, there's no such thing as "the cluster's additionalTrustBundle", there's actually 2 separate trust bundles:

* A configmap in the `openshift-config` namespace called `user-ca-bundle`

* Whatever configmap in the `openshift-config` namespace the `Proxy` CR's `.spec.trustedCA.name` is pointing at (if any), *which might be `user-ca-bundle` itself* but could be technically any other configmap the user created in that namespace

Whatever combination of those configmaps (or lack thereof) the customer has in their original SNO, we need to make sure the new SNO looks the same.

The openshift-installer will _always_ create the `user-ca-bundle` configmap when given install-config `additionalTrustBundle` parameter.

Then whether the installer will make the Proxy CR point at `user-ca-bundle` or not depends on two things:

* Whether the user has a proxy defined in their install-config

* Which `additionalTrustBundle*Policy*` the user chose in their install config

If the user has `additionalTrustBundlePolicy` set to `Proxyonly` (this is the default, so probably most users) then:

* When the user has a proxy set, the Proxy CR will point at `user-ca-bundle`

* When the user doesn't have a proxy set, the proxy CR will *not* point at `user-ca-bundle`

When the user has `additionalTrustBundlePolicy` to `Always` then the proxy CR will point at `user-ca-bundle` regardless of the presence of a proxy in the install-config

None of those cases seem unlikely, so we should know how to handle all of them.

The reason these matter is because cluster-network-operator, which generates the `trusted-ca-bundle` configmap, which is later used and copied by many other operators, will set its value to a concatenation of the value of the `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem` file embedded in the cluster-network-operator container image PLUS whatever configmap the proxy CR is pointing at.

cluster-network-operator* *does not know about `user-ca-bundle` unless the Proxy CR is pointing at it*

However, MCO will set the `ControllerConfig` trust bundle parameter to a concatenation of `user-ca-bundle` and whatever the Proxy CR is pointing at, meaning that *regardless of whether the Proxy CR is pointing at `user-ca-bundle` or not, MCO will pick `user-ca-bundle` up*

This makes `user-ca-bundle` and "whatever configmap the Proxy CR is pointing at" two distinct additional trust bundles, that propagate through the cluster differently, so they have to be handled separately to ensure we support all possible cluster configurations

# Solution / Feature Overview

The recert API would have to be adjusted to receive 3 parameters instead of just one:

* The user-ca-bundle contents (if any)

* The proxy CR trustedCA configmap name (can be empty if the the proxy CR doesn't reference any configmap)

* The proxy CR configmap contents (only needed if the above name is not `"user-ca-bundle"`)

# Implementation Details

The `--additional-trust-bundle` flag is now two separate flags:

* `--user-ca-bundle` which will accept the contents of the `user-ca-bundle` configmap

* `--proxy-trusted-ca-bundle` which will accept the name and the contents of the configmap pointed at by the Proxy CR (colon separated)

Also added missing edits of the trust bundle, one in ControllerConfig under `.spec.additionalTrustBundle` and another in the `user-ca-bundle` configmap itself

# Other Information

## Future work

A corresponding PR in the lifecycle-agent would have to be made to use this new API. There was no existing usage by LCA of the older API so it's fine that's we're breaking it.

## Other changes

* Added new flag, `--postprocess-only`, to skip the crypto stuff. Useful for when you just want to test postprocessing. Not useful in the real world, of course.

* Renamed many instances of static_dir and static_file to the more precise crypto_* or just dir/file phrasing, where applicable

* Deleted old unmaintained `config_example.yaml`, our `./hack/dummy_config.yaml` is good enough.

* Added some logging in places where it makes sense, so it's easier to generally follow what recert is currently doing. Removes some unnecessary logging.

* Renamed the `RecertConfig::new` function to `RecertConfig::load` because it makes more sense. Also refactored some of its code into a separate `ensure_no_cli_args` function

* New rust version introduced a new lint that complains about some of our usages of `clone` in favor of `from_clone`, fixed them

* Removed proxy from `run_seed.sh` because it doesn't work on clusters that don't have a proxy


eranco74:
Failed the initial installation
/retest

omertuc:

omertuc:
/pull-ci-rh-ecosystem-edge-recert-main-e2e-aws-ovn-single-node-recert-serial

omertuc:

PR #160, opened 2024-08-14
MGMT-17267: add etcd encryption support

mresvanis:
This PR adds [OpenShift etcd encryption](https://docs.openshift.com/container-platform/4.16/security/encrypting-etcd.html) support for recert:

- [x] recert now auto-discovers the seed [encryption config](https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/#understanding-the-encryption-at-rest-configuration) and reads OCP encrypted resources using those keys, during `recertify` and `ocp_postprocess`
    - [x] supports AES-GCM (`openssl` doesn't support AES-GCM, so not FIPS compliant anyway, we'll use Rust libs)
    - [x] supports AES-CBC (via `openssl` for FIPS compliance)
- [x] add cmd line args for target's `{kube,openshift,oauth}-apiserver`  [encryption config](https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/#understanding-the-encryption-at-rest-configuration)s to keep target's keys and encrypt all resources with those before committing to etcd
- [x] when seed encryption is enabled and the `{kube,openshift,oauth}-encryption-config` cmd line args have not been provided, generate new ones, encrypt target's resources and put them in etcd and in the filesystem
- [x] ensure no control-plane rollouts occur after recert


src/encrypt.rs
```diff
@@ -0,0 +1,107 @@
...
+                    let prefix = "k8s:enc:aescbc:v1:";
+                    for key in provider.aescbc.context("no AES-CBC provider")?.keys {
+                        let prefix = format!("{}{}:", prefix, key.name);
+                        providers.push(Box::new(aescbc::AesCbc::new(prefix, key.secret)));
+                    }
+                } else if provider.identity.is_some() {
+                    // Nothing to implement
+                } else {
+                    bail!("unsupported provider");
+                }
```
omertuc, line 94: `match` patterns are a good fit here:

```suggestion
                match (provider.aesgcm, provider.aescbc, provider.identity) {
                    (Some(aesgcm), None, None) => {
                        let prefix = "k8s:enc:aesgcm:v1:";
                        for key in aesgcm.keys {
                            let prefix = format!("{}{}:", prefix, key.name);
                            providers.push(Box::new(aesgcm::AesGcm::new(prefix, key.secret)));
                        }
                    }
                    (None, Some(aescbc), None) => {
                        let prefix = "k8s:enc:aescbc:v1:";
                        for key in aescbc.keys {
                            let prefix = format!("{}{}:", prefix, key.name);
                            providers.push(Box::new(aescbc::AesCbc::new(prefix, key.secret)));
                        }
                    }
                    (None, None, Some(_)) => {
                        // Nothing to implement
                    }
                    _ => {
                        bail!("unsupported provider");
                    }
                }

```

src/recert.rs
```diff
@@ -43,15 +83,81 @@ pub(crate) async fn run(recert_config: &RecertConfig, cluster_crypto: &mut Clust
...
 }
 
-async fn get_etcd_endpoint(recert_config: &RecertConfig) -> Result<Arc<InMemoryK8sEtcd>> {
-    let in_memory_etcd_client = Arc::new(InMemoryK8sEtcd::new(match &recert_config.etcd_endpoint {
-        Some(etcd_endpoint) => Some(
-            EtcdClient::connect([etcd_endpoint.as_str()], None)
-                .await
-                .context("connecting to etcd")?,
-        ),
+async fn get_resource_transformers(recert_config: &RecertConfig) -> Result<Option<ResourceTransformers>> {
```
omertuc, line 86: Can be a single match:

```rust
async fn get_resource_transformers(recert_config: &RecertConfig) -> Result<Option<ResourceTransformers>> {
    match &recert_config.encryption_config {
        Some(config_path) => {
            let contents = read_file_to_string(config_path).await.context(format!("reading {config_path}"))?;
            let encryption_config = EncryptionConfiguration::parse_from_file(contents.as_bytes())
                .await
                .context("parsing encryption configuration")?;
            Ok(Some(
                ResourceTransformers::parse_from_encryption_configuration(encryption_config)
                    .await
                    .context("parsing encryption configuration")?,
            ))
        }
        None => Ok(None),
    }
}
```

src/config.rs
```diff
@@ -262,6 +264,10 @@ impl RecertConfig {
             Some(value) => Some(value.as_str().context("etcd_endpoint must be a string")?.to_string()),
             None => None,
         };
+        let encryption_config = match value.remove("encryption_config") {
```
omertuc, line 267: Similarly to our cert/key parameters, we should allow the user to _also_ simply paste the contents of the encryption config directly into the parameter, instead of as a file path - this makes it easier for lifecycle agent to provide recert with this config, without having to worry about annoying temporary files to put the encryption config in (especially annoying because it launches recert inside a container)

src/recert.rs
```diff
@@ -16,7 +22,41 @@ pub(crate) mod timing;
 pub(crate) async fn run(recert_config: &RecertConfig, cluster_crypto: &mut ClusterCryptoObjects) -> Result<RunTimes> {
     ensure_openssl_version().context("checking openssl version compatibility")?;
 
-    let in_memory_etcd_client = get_etcd_endpoint(recert_config).await?;
+    let resource_transformers = get_resource_transformers(recert_config).await?;
+    let mut in_memory_etcd_client = get_etcd_endpoint(recert_config, resource_transformers.clone()).await?;
+
+    if resource_transformers.is_some() {
```
omertuc, line 28: can we tuck this monster `if` somewhere where it doesn't impact the readability of `recert::run`? I don't want people who are trying to understand the general flow of `recert` to be intimidated by all this transofrmer/etcd-encryption stuff which is scary

src/recert.rs
```diff
@@ -16,7 +22,41 @@ pub(crate) mod timing;
...
+        #[allow(clippy::unwrap_used)]
+        let mut resource_to_prefix_transformers: HashMap<_, _> = resource_transformers.unwrap().resource_to_prefix_transformers;
```
omertuc, line 30:
```suggestion
    if let Some(resource_transformers) = resource_transformers {
        let mut resource_to_prefix_transformers: HashMap<_, _> = resource_transformers.resource_to_prefix_transformers;
```

src/recert.rs
```diff
@@ -16,7 +22,41 @@ pub(crate) mod timing;
...
+
+        let apiserver_location = K8sResourceLocation::new(Some("openshift-apiserver"), "Secret", "encryption-config", "v1");
+        if let Some(apiserver_resource_transformers) =
+            get_openshift_resource_transformers(&in_memory_etcd_client, &apiserver_location).await?
+        {
+            resource_to_prefix_transformers = resource_to_prefix_transformers
+                .into_iter()
+                .chain(apiserver_resource_transformers.resource_to_prefix_transformers)
```
omertuc, line 38: Since the hashmap is already `mut`, you can `.extend` it. If you want to use `chain` (which I prefer), it's better it you remove `mut` and redeclare (shadowing) `resource_to_prefix_transformers` with another `let resource_to_prefix_transformers = ...`

omertuc (changes requested):
Very cool, mostly nits and one very annoying request - to switch to `openssl` and drop the dependencies, for FIPS reasons

omertuc:
Is there any reason why we ask the user to provide us with the encryption config for the main API server yet we fetch the encryption config for openshift-apiserver and the oauth-apiserver on our own?

Cargo.toml
```diff
@@ -48,6 +48,10 @@ prost = "0.12.1"
 log = "0.4.20"
 itertools = "0.12.1"
 serial_test = "3.0.0"
+aes-gcm = "0.10.3"
```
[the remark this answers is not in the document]
mresvanis: Very good point and thank you for the validation, I was thinking about this just today when testing this draft e2e. I'll move forward with using `openssl` instead (tbh I was't 100% sure about those libs either, I would have had to do some more digging to ensure they were well-maintained etc etc) :+1:

src/encrypt.rs
```diff
@@ -0,0 +1,107 @@
...
+
+        ensure!(config.kind == KIND, format!("kind should equal {}", KIND));
+        ensure!(config.apiVersion == API_VERSION, format!("apiVersion should equal {}", API_VERSION));
+
+        Ok(config)
+    }
+}
+
+impl ResourceTransformers {
+    #[allow(clippy::if_same_then_else)]
```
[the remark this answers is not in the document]
mresvanis: Nice catch, you're right! Before implementing AES-CBC there were 2 conditions with empty implementations and this clippy rule was triggered. I'll remove it :+1:

src/config.rs
mresvanis: Good point, just today I tried adding this arg to LCA and it was indeed very very annoying :) I'll refactor this, thanks for the validation :+1:

src/recert.rs
mresvanis: oh nice, I agree with avoiding `mut`, let's use `shadowing` instead. I'll update this  :+1:

mresvanis:
> What stops us from simply fetching the encryption config for kube-apiserver, instead of asking the user to provide us with this config?

The `openshift-kube-apiserver/encryption-config` is a `Secret`, which is encrypted. That leads us to the filesystem encryption-config file (under `/etc/kubernetes/static-pod-resources/kube-apiserver-pod-<latest>/secrets/encryption-config/encryption-config`) to fetch all the encryption details for the kube-apiserver encrypted resources (i.e. `Secrets` and `ConfigMaps`) from. Because it's a file and we're using recert mostly in a container, where we mount everything we need from the filesystem, I thought it best to have a path argument (which I will refactor to also accept the content of that file, as suggested [here](https://github.com/rh-ecosystem-edge/recert/pull/160#discussion_r1719929935)). Does this make sense?

src/recert.rs
mresvanis: I've hidden the "storage transformers" logic under a `setup_etcd_client()` function in `recert::run`. Future readers can dig into that if interested, otherwise I think it's clear that this function does exactly what its name suggests :) 

Let me know if you were thinking of some other alternative :pray:

src/config.rs
mresvanis: Updated.

src/config/cli.rs
```diff
@@ -18,6 +19,22 @@ pub(crate) struct Cli {
...
+    // - Routes encrypted by the openshift-kube-apiserver
+    // - oauth {access,authorize} tokens encrypted by the openshift-oauth-apiserver
+    // The secrets/openshift-kube-apiserver/encryption-config and the secrets/openshift-oauth-apiserver/encryption-config
+    // are encrypted by the kube-apiserver and we need to decrypt those first in order to decrypt Routes and oauth token resources.
+    // Thus we resort to the kube-apiserver encryption-config file (i.e. /etc/kubernetes/static-pod-resources/kube-apiserver-pod-<latest>/secrets/encryption-config/encryption-config)
+    // to fetch the encryption details for the kube-apiserver encrypted resources (i.e. Secrets and ConfigMaps) first.
+    /// Kubernetes EncryptionConfiguration resource in JSON formatted string or path to the
+    /// respective file under /etc/kubernetes/static-pod-resources/kube-apiserver-pod-<latest>/secrets/encryption-config/encryption-config.
+    /// When specified, recert can be executed on OpenShift clusters with encryption at rest enabled.
+    /// At the end of its execution it will generate new encryption keys and use them to encrypt the
```
omertuc, line 33: Do we always want to re-encrypt? Probably yes, we never want the seed encryption keys. But do we always want to re-encrypt with random keys? Not so sure. What if some customers care about the particular keys with which their etcd database is encrypted? In that case, we would have to preserve those keys from the cluster being upgraded into the new cluster, so the new encryption keys should be configurable.

What do you think?

mresvanis: Valid point! The [OpenShift etcd encryption docs](https://docs.openshift.com/container-platform/4.16/security/encrypting-etcd.html) do not address anything regarding the encryption keys and nor do the [OpenShift etcd backup ones](https://docs.openshift.com/container-platform/4.16/backup_and_restore/control_plane_backup_and_restore/backing-up-etcd.html). When taking an etcd backup, alonside the etcd db there is a tarball with the `static-pod-resources` folder (which includes the `encryption-config` at that point in time).

```sh
$ ls -l
total 103176
-rw-------. 1 root root 105570336 Sep 23 12:57 snapshot_2024-09-23_125752.db
-rw-------. 1 root root     75121 Sep 23 12:57 static_kuberesources_2024-09-23_125752.tar.gz
```

I don't think users care about whether their encryption keys change or not. But I do see the point in implementing this feature like any such IBU feature, i.e. to keep the target cluster's encryption keys intact and add a boolean cmd line argument to forcefully rotate them when that's set (e.g. `--rotate-encryption-keys`).

That said, I would:

1. add a `--seed-encryption-config` cmd line arg and use that to decrypt
2. keep the current one, i.e. `--encryption-config` and use that one to encrypt and put it in the filesystem and etcd
3. implement the forceful encryption keys rotation in a subsequent PR

WDYT?

Cargo.toml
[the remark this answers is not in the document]
mresvanis: I've [replaced](https://github.com/rh-ecosystem-edge/recert/pull/160/files#diff-8c4912bab47ef7cc1eddfef6a5ce4017becc2dd2f69bf098e09f8d48dbfcdc19R31-R185) Rust libs with `openssl` for the AES-CBC mode and it doesn't seem to be that slower (or maybe even has the same performance, I'll get some numbers to support this). 

But we can't do the same for AES-GCM, as per [this](https://github.com/openssl/openssl/blob/260ecea0d4e46d63464636405f9925ef65d0747e/doc/man1/openssl-enc.pod.in#L299-L314) openssl doc (I tested it just to be sure that this is actually the case). TL:DR openssl doesn't and won't support AEAD ciphers, because it can't validate the authentication tag until the very end of the stream, at which point it can't rollback all the data that it has already deciphered. 

I'm now looking into other FIPS compliant options. Please let me know if you have any ideas.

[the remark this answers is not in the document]
mresvanis: When looking into FIPS, OpenShift etcd encryption and openssl, I'm assuming that: 

since Go libs are not FIPS compliant and are replaced by openssl and openssl doesn't support AES-GCM, the OCP etcd AES-GCM encryption is by default NOT FIPS compliant. 

So we can use the AES Rust libs to support AES-GCM etcd encryption, but we should document that using AES-GCM is NOT FIPS compliant. WDYT?

src/config/cli.rs
mresvanis: What I suggested above is not going to be enough, as we need 3 encryption configs (`kube-apiserver`, `openshift-apiserver`, `oauth-apiserver`), so let's scratch that. I'll come up with a couple of alternatives.

mresvanis: What I came up with (and started working on) is the following:

- auto-discover if seed encryption is enabled and the encryption type used (by reading the `apiserver/cluster`'s  `spec.encryption.type` field), then load the seed `encryption-config`s from the filesystem and etcd
- add 3 optional recert config args -> `{kube,openshift,oauth}-encryption-config`. When provided recert will use those to encrypt all resources to be encrypted, otherwise it will generate new `encryption-config`s
- put the `encryption-config`s used (either the user provided or the newly generated ones) in etcd and the filesystem

The above assumes that recert will  try to encrypt only when the seed is encrypted. In the case of the seed not being encrypted, recert will reject the `encryption-config` args. Does this make sense?

omertuc:
Create a ticket for LCA health-check to make sure no encryption is in-progress

omertuc:
Meeting note: Don't use rust rand, use openssl rand

omertuc:
Meeting note: If openshift ever adds a component like openshift-apiserver or oauth, can we assert and have a clear error for it?

omertuc:
Meeting note: assert no kinds overlap across api/ocp-api/oauth

omertuc:
Meeting note: encryption_config::etcd_rename::reencyrpt_resources should not hardcode the chain kinds

Cargo.toml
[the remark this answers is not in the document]
mresvanis: Resolved offline that this makes sense :+1:

src/recert.rs
```diff
@@ -44,18 +53,299 @@ pub(crate) async fn run(recert_config: &RecertConfig, cluster_crypto: &mut Clust
...
+        .as_str()
+        .context("spec.encryption.type not a string")?;
+
+    Ok(encryption_type.to_string())
+}
+
+async fn setup_etcd_client(recert_config: &RecertConfig) -> Result<(Arc<InMemoryK8sEtcd>, Option<EncryptionCustomizations>)> {
+    let mut in_memory_etcd_client = get_etcd_endpoint(recert_config, None, None).await?;
 
     log::info!("Connected to etcd");
```
omertuc, line 123: This log now happens before we even connect, so it's a bit misleadingingly successful

omertuc (changes requested):
Amazing work, mostly superficial comments

src/encrypt/aescbc.rs
```diff
@@ -0,0 +1,200 @@
...
+        stdout_result?;
+        stderr_result?;
+
+        let status = command.wait().await.context("failed to wait for openssl")?;
+
+        if status.success() {
+            Ok(plaintext)
+        } else {
+            bail!(format!("AES-CBC decryption error: {}", String::from_utf8_lossy(&stderr)));
+        }
```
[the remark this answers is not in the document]
mresvanis: Awesome simplification, I was sure it wasn't as readable as it can be :)

But it just blocks, any idea what are we missing compared to the previous version?

src/recert.rs
```diff
@@ -44,18 +53,299 @@ pub(crate) async fn run(recert_config: &RecertConfig, cluster_crypto: &mut Clust
...
+    let k8s_location = K8sResourceLocation::new(None, "APIServer", "cluster", "config.openshift.io/v1");
+
+    let cluster = get_etcd_json(in_memory_etcd_client, &k8s_location)
+        .await?
+        .context(format!("could not get {}", k8s_location))?;
+
+    if let Some(encryption_type) = cluster.pointer("/spec/encryption/type") {
+        let encryption_type = encryption_type.as_str().context("spec.encryption.type not a string")?;
+
+        Ok(encryption_type == "aescbc" || encryption_type == "aesgcm")
```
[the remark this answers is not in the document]
mresvanis: Of course, thanks for catching that. I have updated it with the respective `ensure!`.

src/encrypt/aescbc.rs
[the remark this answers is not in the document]
omertuc: Hmm, don't know

src/recert.rs
```diff
@@ -44,18 +53,299 @@ pub(crate) async fn run(recert_config: &RecertConfig, cluster_crypto: &mut Clust
...
-async fn get_etcd_endpoint(recert_config: &RecertConfig) -> Result<Arc<InMemoryK8sEtcd>> {
-    let in_memory_etcd_client = Arc::new(InMemoryK8sEtcd::new(match &recert_config.etcd_endpoint {
-        Some(etcd_endpoint) => Some(
-            EtcdClient::connect([etcd_endpoint.as_str()], None)
-                .await
-                .context("connecting to etcd")?,
-        ),
-        None => None,
-    }));
+async fn is_encryption_enabled(in_memory_etcd_client: &Arc<InMemoryK8sEtcd>) -> Result<bool> {
```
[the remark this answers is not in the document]
mresvanis: Great point, I have put all `recert.rs` encryption related code to its own module under `recert/encrypt_utils.rs`. WDYT?

src/encrypt/aescbc.rs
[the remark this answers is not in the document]
omertuc:
```rust
    let (mut command_stdin, mut command_stdout, mut command_stderr) = (
        command.stdin.take().context("failed to take stdin")?,
        command.stdout.take().context("failed to take stdout")?,
        command.stderr.take().context("failed to take stderr")?,
    );

    command_stdin.write_all(data).await.context("failed to write stdin")?;
    drop(command_stdin);

    command_stdout.read_to_end(&mut plaintext).await.context("failed to read stdout")?;
    command_stderr.read_to_end(&mut stderr).await.context("failed to read stderr")?;
```

This should be good enough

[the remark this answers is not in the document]
mresvanis: This still blocks at some point, I think if we read concurrently `stdout` and `stderr` this issue will be solved.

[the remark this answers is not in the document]
mresvanis: > This still blocks at some point, I think if we read concurrently stdout and stderr this issue will be solved.

No, it doesn't actually.

src/encrypt.rs
```diff
@@ -0,0 +1,220 @@
...
+                res.providers = vec![res.providers[0].clone()]
+            }
+        }
+    }
+}
+
+async fn generate_key() -> Result<String> {
+    let output = Command::new("openssl")
+        .arg("rand")
+        .arg("-base64")
```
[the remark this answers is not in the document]
mresvanis: Great point! This makes total sense, I have updated the PR accordingly. PTAL whenever possible.

src/encrypt.rs
```diff
@@ -0,0 +1,221 @@
...
+        .arg("32")
+        .output()
+        .await
+        .context("failed to run openssl command")?;
+
+    ensure!(
+        output.status.success(),
+        format!("openssl command failed with status: {}", output.status)
+    );
+
```
omertuc, line 175: We should probably `ensure!` we're reading exactly 32 bytes and not some random CLI junk

Also move `32` (also used above) into a constant (const AES_KEY_SIZE 256, AES_KEY_SIZE_BYTES = AES_KEY_SIZE / 8) and use it in the ensure

omertuc (changes requested):
Thanks, small change

src/encrypt.rs
mresvanis: Once more, this makes total sense. Is this what you had in mind?

src/encrypt/aescbc.rs
```diff
@@ -0,0 +1,200 @@
...
+            .arg(&key_hex)
+            .arg("-iv")
+            .arg(&iv_hex)
+            .stdin(Stdio::piped())
+            .stdout(Stdio::piped())
+            .stderr(Stdio::piped())
+            .spawn()
+            .context("failed to spawn openssl")?;
+
+        let stdin_future = async {
```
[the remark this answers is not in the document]
mresvanis: Nice, I added this function as suggested, it looks a lot better now. PTAL whevener possible.

src/encrypt.rs
```diff
@@ -0,0 +1,229 @@
...
+pub(crate) mod aescbc;
+pub(crate) mod aesgcm;
+pub(crate) mod transformer;
+
+const API_VERSION: &str = "apiserver.config.k8s.io/v1";
+const KIND: &str = "EncryptionConfiguration";
+
+const AES_KEY_SIZE: i32 = 256;
+const AES_KEY_SIZE_BYTES: i32 = AES_KEY_SIZE / 8;
```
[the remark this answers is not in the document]
mresvanis: Thanks for catching that and even more for the "ramblings"! I've updated it to `usize`, not sure why that went in as `i32`, in other cases (e.g. the `BLOCK_SIZE` const) I've properly used `usize`. 

PS. your example feels too specific to be imaginary :D

omertuc:
Compare with https://github.com/rh-ecosystem-edge/recert/pull/168

mresvanis:
/override e2e-aws-ovn-single-node-recert-parallel e2e-aws-ovn-single-node-recert-serial

Those are failing and need debugging even without this change.

mresvanis:
/override ci/prow/e2e-aws-ovn-single-node-recert-parallel ci/prow/e2e-aws-ovn-single-node-recert-serial

PR #161, opened 2024-09-11
Setting new ip to etcd member

tsorya:
This commit implements logic from lca https://github.com/openshift-kni/lifecycle-agent/blob/main/lca-cli/postpivot/postpivot.go#L358

As we moved all the ip change logic into recert we should handle this part here too.

This code updates existing member with new ip

### DIFF
--- a/src/k8s_etcd.rs
+++ b/src/k8s_etcd.rs
@@ -1,6 +1,6 @@
 use crate::cluster_crypto::locations::K8sResourceLocation;
 use crate::etcd_encoding;
-use anyhow::{bail, Context, Result};
+use anyhow::{bail, ensure, Context, Result};
 use etcd_client::{Client as EtcdClient, GetOptions};
 use futures_util::future::join_all;
 use serde_json::Value;
@@ -174,6 +174,36 @@ impl InMemoryK8sEtcd {
     pub(crate) async fn delete(&self, key: &str) -> Result<()> {
         self.etcd_keyvalue_hashmap.lock().await.remove(key);
         self.deleted_keys.lock().await.insert(key.to_string());
+        Ok(())
+    }
+
+    pub(crate) async fn update_member(&self, value: String) -> Result<()> {
+        let etcd_client = self.etcd_client.as_ref().context("etcd client not configured")?;
+
+        let members_list = etcd_client
+            .cluster_client()
+            .member_list()
+            .await
+            .context("listing etcd members list")?;
+
+        let members = members_list.members();
+
+        ensure!(
+            members.len() == 1,
+            "single-node must have exactly one etcd member, found {}",
+            members.len()
+        );
+
+        ensure!(
+            !etcd_client
+                .cluster_client()
+                .member_update(members[0].id(), vec![value.clone()])
+                .await
+                .context("updating etcd member")?
+                .members()
+                .is_empty(),
+            "no members in update response"
+        );
         Ok(())
     }
 }
--- a/src/ocp_postprocess/ip_rename/etcd_rename.rs
+++ b/src/ocp_postprocess/ip_rename/etcd_rename.rs
@@ -6,6 +6,7 @@ use crate::{
 use anyhow::{bail, ensure, Context, Result};
 use futures_util::future::join_all;
 use serde_json::Value;
+use std::net::Ipv6Addr;
 use std::sync::Arc;
 
 pub(crate) async fn fix_openshift_apiserver_configmap(etcd_client: &Arc<InMemoryK8sEtcd>, ip: &str) -> Result<String> {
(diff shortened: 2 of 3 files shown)


src/k8s_etcd.rs
```diff
@@ -176,6 +176,34 @@ impl InMemoryK8sEtcd {
         self.deleted_keys.lock().await.insert(key.to_string());
         Ok(())
     }
+
+    pub(crate) async fn update_member(&self, value: String) -> Result<()> {
+        let etcd_client = match &self.etcd_client {
+            Some(etcd_client) => etcd_client,
+            None => bail!("etcd client not configured"),
+        };
```
omertuc, line 184: `context` on an `Option` turns it into a `Result`, so we can write the error directly on it then `?` to get equivalent code:

```suggestion
        let etcd_client = self.etcd_client.context("etcd client not configured")?;
```

src/k8s_etcd.rs
```diff
@@ -176,6 +176,34 @@ impl InMemoryK8sEtcd {
...
+
+        let members_list = etcd_client
+            .cluster_client()
+            .member_list()
+            .await
+            .context("during etcd members list")?;
```
omertuc, line 190: `.context` on `Result` should describe what we are doing, so `during` is a bit too verbose

```suggestion
            .context("listing etcd members")?;
```

src/k8s_etcd.rs
```diff
@@ -176,6 +176,34 @@ impl InMemoryK8sEtcd {
...
+
```
omertuc, line 191: We're assuming the first member is the only one, so might as well add an assertion to make sure nothing funny is happening: 

```suggestion
ensure!(members_list.members().len() == 1, "single-node must have exactly one etcd member, found {}", members_list.members().len());
```

src/k8s_etcd.rs
```diff
@@ -176,6 +176,34 @@ impl InMemoryK8sEtcd {
...
+            let mut vec: Vec<String> = Vec::new();
+            vec.push(value.clone());
+            let update_response = etcd_client
+                .cluster_client()
+                .member_update(member_id, vec)
+                .await
+                .context("during etcd member update")?;
+            if update_response.members().is_empty() {
+                bail!("no members in update response");
+            }
```
omertuc, line 203: Can be simplified

```suggestion
            ensure!(!etcd_client
                .cluster_client()
                .member_update(member.id(), vec![value.clone()])
                .await
                .context("updating etcd member")?.members().is_empty(), "no members in update response")
```

src/ocp_postprocess/ip_rename/etcd_rename.rs
```diff
@@ -448,6 +449,16 @@ pub(crate) async fn fix_networks_cluster(etcd_client: &Arc<InMemoryK8sEtcd>, ip:
     Ok(())
 }
 
+pub(crate) async fn fix_etcd_member(etcd_client: &Arc<InMemoryK8sEtcd>, ip: &str) -> Result<()> {
+    let mut update = format!("https://{}:2380", ip).to_string();
+    if !ip.parse::<Ipv6Addr>().is_err() {
```
omertuc, line 454: positive easier to read than double negative

```suggestion
    if ip.parse::<Ipv6Addr>().is_ok() {
```

omertuc (changes requested):
Cool, thanks. Added some rust syntax suggestions but otherwise looks good

src/k8s_etcd.rs
```diff
@@ -176,6 +176,36 @@ impl InMemoryK8sEtcd {
         self.deleted_keys.lock().await.insert(key.to_string());
         Ok(())
     }
+
+    pub(crate) async fn update_member(&self, value: String) -> Result<()> {
+        let etcd_client = self.etcd_client.clone().context("etcd client not configured")?;
```
tsorya, line 181: needed to set clone as without clone it was failing 
` move occurs because `self.etcd_client` has type `std::option::Option<Arc<Client>>`, which does not implement the `Copy` trait`

omertuc: We can just borrow the Arc instead of cloning the option

```suggestion
        let etcd_client = self.etcd_client.as_ref().context("etcd client not configured")?;
```

This should work as well

src/k8s_etcd.rs
```diff
@@ -176,6 +176,36 @@ impl InMemoryK8sEtcd {
...
+                !etcd_client
+                    .cluster_client()
+                    .member_update(member.id(), vec![value.clone()])
+                    .await
+                    .context("updating etcd member")?
+                    .members()
+                    .is_empty(),
+                "no members in update response"
+            )
+        };
```
omertuc, line 206: With the above `ensure!` we can safely remove the `if` statement:

```suggestion
        ensure!(
            !etcd_client
                .cluster_client()
                .member_update(members[0].id(), vec![value.clone()])
                .await
                .context("updating etcd member")?
                .members()
                .is_empty(),
            "no members in update response"
        )
```

omertuc: But since every time we call `members()` it's an API call to etcd maybe it's best to save the members list in a variable. The full code should look like:

```rust
        let members = members_list.members();

        ensure!(
            members.len() == 1,
            "single-node must have exactly one etcd member, found {}",
            members.len()
        );

        ensure!(
            !etcd_client
                .cluster_client()
                .member_update(members[0].id(), vec![value.clone()])
                .await
                .context("updating etcd member")?
                .members()
                .is_empty(),
            "no members in update response"
        );
```

PR #162, opened 2024-09-12
Set CVO available condition to False

mresvanis:
This PR updates the `ClusterVersion`'s status condition of type `Available` to `status: "False"`, along with its `message: "Cluster version status unknown"`, in the OCP post-processing step of recert.

/cc @tsorya

### DIFF
--- a/src/ocp_postprocess.rs
+++ b/src/ocp_postprocess.rs
@@ -59,6 +59,16 @@ pub(crate) async fn ocp_postprocess(
 
     run_cluster_customizations(cluster_customizations, in_memory_etcd_client).await?;
 
+    // When OpenShift pods/containers start, CVO is still stuck on its last known status and it
+    // takes a couple of minutes for it to update its status to the current cluster conditions. When
+    // its last known status is set to Available=True, it incorrectly shows that OpenShift is
+    // stabilized while it's not. Since we want to watch the CVO status to signal the cluster
+    // installation/update completion, this flapping CVO status makes it difficult. By setting CVO's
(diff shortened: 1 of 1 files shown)


src/ocp_postprocess.rs
```diff
@@ -59,6 +59,8 @@ pub(crate) async fn ocp_postprocess(
 
     run_cluster_customizations(cluster_customizations, in_memory_etcd_client).await?;
 
+    set_cluster_version_available_false(in_memory_etcd_client).await?;
```
omertuc, line 62:
```suggestion
    // We do this because ...
    set_cluster_version_available_false(in_memory_etcd_client).await?;
```

omertuc (changes requested):

src/ocp_postprocess.rs
mresvanis: Good point, I'll talk with Igal to verify the reason and update this change with a proper comment.

mresvanis:
/hold 

This change needs to be documented with the respective comment.

src/ocp_postprocess.rs
tsorya: Currently it is very difficult to say when we finish installation and when data is not stale. 
Right now looking on cvo status does nothing as cvo doesn't even update this condition while starting in ibi/ibu.
This change allows us to monitor cluster installation with regular tools like looking on cvo available condition that will be set back to True when cvo will validate that everything is ok

mresvanis: @omertuc updated, PTAL.

PR #165, opened 2024-09-17
Add etcd defragmentaion option

tsorya:
in order to remove defragmentation code from lca , adding it to recert with flag, so that we would run it only on demand


src/config/cli.rs
```diff
@@ -192,4 +192,8 @@ pub(crate) struct Cli {
     /// those intentionally expired dates.
     #[clap(long, groups = &["dry", "expiration"])]
     pub(crate) force_expire: bool,
+
+    /// Run etcd defragment command after recertification
+    #[clap(long)]
```
omertuc, line 197: Use `groups` or `group` to ensure exclusivity with dry run

omertuc:

PR #174, opened 2025-02-20
jwt: calculate kid from public key

omertuc:
# Background

Recert resigns JWTs found in the cluster. The kid field in the Kubernetes JWT header is a Base64URL encoded hash of the public key used to sign the JWT. The kid used by verifiers (e.g. kube-apiserver) to quickly identify the key used to sign the JWT, instead of brute-forcing.

# Issue

See https://issues.redhat.com/browse/OCPBUGS-49972

The current implementation uses the SHA256 hash of the private key to calculate the kid. This is inconsistent with [the Kubernetes implementation](https://github.com/openshift/kubernetes/blob/6fdacf04117cef54a0babd0945e8ef87d0f9461d/pkg/serviceaccount/jwt.go#L92-L112) which uses the public key to calculate the kid.

In recent versions of Kubernetes, a kid mismatch leads to "unauthorized" errors, which eventually leads to a collapse of the cluster.

# Solution

This commit changes the kid calculation to use the SHA256 hash of the public key instead of the private key, so that it becomes more consistent with Kubernetes

# Backwards compatibility

Seeds processed with older versions of recert that don't contain this fix should still work if "rejuvenated" by newer versions of recert which do contain this fix. This is because the new version will simply overwrite the `kid` field with the correct value.

## Summary by CodeRabbit

- **New Features**
	- Introduced a method for generating a "kid" field for JWT headers, enhancing token identification.
	- Added a test module to validate the new JWT key ID functionality.

- **Refactor**
	- Streamlined import statements for better code clarity.
	- Improved variable naming for better readability in the resigning logic.
	- Updated error handling messages for clearer feedback.

### DIFF
--- a/src/cluster_crypto/crypto_utils.rs
+++ b/src/cluster_crypto/crypto_utils.rs
@@ -1,7 +1,11 @@
 use super::certificate;
 use anyhow::ensure;
 use anyhow::{bail, Context, Result};
-use base64::{engine::general_purpose::STANDARD as base64_standard, Engine as _};
+use base64::{
+    engine::general_purpose::{STANDARD as base64_standard, URL_SAFE_NO_PAD as base64_url},
+    Engine as _,
+};
+use bcder::BitString;
 use bcder::{encode::Values, Mode};
 use pkcs1::DecodeRsaPrivateKey;
 use rsa::{self, pkcs8::EncodePrivateKey, RsaPrivateKey};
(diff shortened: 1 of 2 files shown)


src/cluster_crypto/crypto_utils.rs
```diff
@@ -275,3 +301,101 @@ pub(crate) fn ensure_openssl_version() -> Result<()> {
...
+            hasher.Write(publicKeyDERBytes)
+            publicKeyDERHash := hasher.Sum(nil)
+
+            keyID := base64.RawURLEncoding.EncodeToString(publicKeyDERHash)
+
+            fmt.Println(keyID)
+        }
+        */
+    }
+}
```
eranco74, line 401: ?

eranco74 (commented):
/lgtm

src/cluster_crypto/crypto_utils.rs
omertuc: >        // Got this one from the Go code below

omertuc:

PR #176, opened 2025-02-25
Print cert locations when re-signing fails

mresvanis:
This change adds the certificate locations to the `re_sign_cert(...)`'s error message to have more information about the certificate that caused the error.

For example, we currently have the following error message:

```text
Caused by:
    0: processing discovered objects
    1: regenerating crypto
    2: re-signing cert with subject CN=172.30.0.1
    3: could not find matching akid key identifier in chain
```

with this change we will have:

```text
Caused by:
    0: processing discovered objects
    1: regenerating crypto
    2: re-signing cert at locations [file:/kubelet/pods/<some-pod>/volumes/kubernetes.io~projected/etc-kubernetes-pki/etcd/server.crt:in PEM bundle at index 0] with subject CN=172.30.0.1
```

## Summary by CodeRabbit

- **Refactor**
	- Enhanced certificate error messages by including additional location details alongside subject information for improved troubleshooting context.

### DIFF
--- a/src/cluster_crypto/cert_key_pair.rs
+++ b/src/cluster_crypto/cert_key_pair.rs
@@ -135,7 +135,7 @@ impl CertKeyPair {
     /// # Errors
     ///
     /// This function will return an error if .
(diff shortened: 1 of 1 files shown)


mresvanis:
@coderabbitai full review

PR #177, opened 2025-03-04
OCPBUGS-52178: Filter out OCP certs from the external ones

mresvanis:
When using RHACM to deploy disconnected IBU seed SNO clusters, the trusted-ca-bundle, which should only contain external to OCP CA certificates, can contain also OCP internal CA certificates. E.g.:

- internal-loadbalancer-serving
- service-network-serving-cert
- localhost-serving-cert
- external-loadbalancer-serving

This change filters out all certs found in there with their subject containing openshift as the organisation unit (i.e. OU=openshift), as they are not external certs and should not be ignored by recert when regenerating and re-signing OCP certificates.

## Summary by CodeRabbit

- **New Features**
  - Enhanced the certificate discovery process by introducing an improved filtering step that automatically excludes OpenShift-related certificates based on their subject names.
  - Added a new helper function to identify OpenShift-related certificates.
- **Tests**
  - Introduced unit tests to verify the behavior of the new helper function against various subject name inputs.

### DIFF
--- a/src/cluster_crypto/scanning/external_certs.rs
+++ b/src/cluster_crypto/scanning/external_certs.rs
@@ -1,11 +1,27 @@
 use super::super::locations::K8sResourceLocation;
 use crate::k8s_etcd::get_etcd_json;
 use crate::k8s_etcd::InMemoryK8sEtcd;
-use anyhow::{bail, Context, Result};
+use anyhow::{Context, Result};
 use itertools::Itertools;
 use std::collections::HashSet;
 use std::sync::Arc;
 use x509_certificate::X509Certificate;
+
+/// Patterns used to identify OpenShift-specific certificates that should be
+/// filtered out from external certificates.
+const OPENSHIFT_CN_PATTERNS: &[&str] = &[
+    "ou=openshift",
+    "cn=ingress-operator@",
+    "cn=openshift-kube-apiserver-operator_localhost-recovery-serving-signer@",
+];
+
+/// Determines if a certificate is an OpenShift internal certificate by
+/// checking if its subject contains any of the OpenShift-specific patterns.
+/// The check is case-insensitive.
+fn is_openshift_certificate(subject: &str) -> bool {
+    let subject_lower = subject.to_lowercase();
+    OPENSHIFT_CN_PATTERNS.iter().any(|pattern| subject_lower.contains(*pattern))
+}
 
 pub(crate) async fn discover_external_certs(in_memory_etcd_client: Arc<InMemoryK8sEtcd>) -> Result<HashSet<String>> {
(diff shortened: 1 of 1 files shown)


mresvanis:
@coderabbitai review

mresvanis:
@coderabbitai review

mresvanis:
Although this change filters out some of the OCP internal certs found in the user CA bundle, I believe there are others. E.g.: 

```log
2025-03-04 17:13:27 - TRACE - src/cluster_crypto/scanning/external_certs.rs:43: Found external certificate: CN=openshift-kube-apiserver-operator_localhost-recovery-serving-signer@1722437520
2025-03-04 17:13:27 - TRACE - src/cluster_crypto/scanning/external_certs.rs:43: Found external certificate: CN=*.apps.seed.mylab.com
2025-03-04 17:13:27 - TRACE - src/cluster_crypto/scanning/external_certs.rs:43: Found external certificate: CN=ingress-operator@1722437639
```

mresvanis:
@coderabbitai full review

mresvanis:
@coderabbitai full review

mresvanis:
@coderabbitai full review

omertuc:
I'm not sure I'm comfortable with this change. Some cluster X might trust a certificate from a different cluster Y, which would make said certificate external to X but still, that certificate could have `openshift` markings in it, which would make recert consider it internal and thus regenerate it, breaking the connectivity between the clusters.

But I'm only complaining without offering a solution because I don't quite have one yet. We need to figure out a different heuristic to determine external-ness. 

We probably need to discuss it

mresvanis:
> I'm not sure I'm comfortable with this change.

This change is here mostly for us to have a starting point for this discussion. As it is I wouldn't merge it either, we need to go for a more thorough solution if at all.

> Some cluster X might trust a certificate from a different cluster Y, which would make said certificate external to X but still, that certificate could have openshift markings in its name, which would make recert consider it internal and thus regenerate it, breaking the connectivity between the clusters (e.g. hub-spoke relationship?)

I don't have the details about the use-cases on hub-spoke setups, so this could very well be the case. (I mean RHACM uses mTLS in some parts and the klusterlet-agent with a kubeconfig to connect to the hub, but that's as far as I can go for now). 

That's why I started thinking about a more robust solution. We should keep treating all `trusted-ca-bundle` certs as external, but if one of them can be found in a location that contains internal to this cluster certs (e.g. `/etc/kubernetes/static-pod-resources/kube-apiserver-certs`) then we pull it out and treat it as internal.

 > If what I'm saying makes sense, we need to figure out a different heuristic to determine external-ness.
 
It makes sense and perhaps, if that simplifies it, I would reverse it to determine internal-ness :) I can propose another solution and of course shoot if you already have something in mind.

mresvanis:
In favor of #188

PR #207, opened 2025-04-25
[release-4.19] Update Konflux target branch to release-4.19

fontivan:
Update the branch Konflux is checking against

### DIFF
--- a/.tekton/recert-4-19-push.yaml
+++ b/.tekton/recert-4-19-push.yaml
@@ -8,7 +8,7 @@ metadata:
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/cancel-in-progress: "false"
     pipelinesascode.tekton.dev/max-keep-runs: "3"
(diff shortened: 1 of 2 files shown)


.tekton/recert-4-19-push.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/cancel-in-progress: "false"
     pipelinesascode.tekton.dev/max-keep-runs: "3"
-    pipelinesascode.tekton.dev/on-cel-expression: event == "push" && target_branch == "main" && files.all != ["konflux_clusterserviceversion_overlay.data"]
```
rauhersu, line 11: I believe we don't need neither that '.data' file nor that logic present on the recert repo.

fontivan: I'm actually not sure how that interacts now since the bundle is the other repo, I may leave this for now and revisit it later

PR #332, opened 2025-06-20
Pin rust version to the version available in rhel9

fontivan:
- We are using rhel9 images for production, so we want to make sure the rust version used here is compatible
- rhel9 currently uses rust 1.75


Cargo.toml
```diff
@@ -1,5 +1,7 @@
 [package]
 name = "recert"
+# we need to compile with a rust version that is compatible with rhel9
+rust-version = "1.75"
```
omertuc, line 4: Can you explain why this is necessary? If the dockerfile uses 1.87, why would we need to specify 1.75 as the minimum version? 

I mean I think it's good in general but maybe our comment should be more specific about why this particular version of rust was decided as the minimum?

fontivan: I've updated the comment here to be more descriptive

PR #1493, opened 2026-05-11
Support new library-go SKID calculation method

omertuc:
library-go always had an awkward incorrect method to calculate SKIDs for RSA keys. Recently they added support for calculating based on RFC7093 but that implementation is also wrong in a similar way to the original wrong one.

Recert needs to be aware of this new wrong calculation method so this PR adds it to recert

Also added comprehensive test data to ensure I didn't mess anything up during the light refactoring. Sorry for the certificate binary blobs, they were taken from a recert summary that I have no control over and generating them here would beat the point (as we're trying to ensure we're compatible with silly OCP operators). Added a simple test to check the self-sign signature of the binary blob certs to somewhat ensure they're not being arbitrarily modified


danmanor (approved):

omertuc:

omertuc:

omertuc:

omertuc:
/override pull-ci-rh-ecosystem-edge-recert-main-baremetalds-sno-recert-cluster-rename

PR #1550, opened 2026-05-21
Fix etcd static pod hostname renaming for non-shell containers

omertuc:
Required due to https://github.com/openshift/cluster-etcd-operator/pull/1606

The `etcd-ensure-env-vars` init container changed from an inline bash script to a Go binary (cluster-etcd-operator ensure-env), which caused `fix_etcd_static_pod_container` to skip renaming hostname references in command args. The function relied on finding a `"#!/bin/sh\n"` prefix to identify args containing hostname-dependent strings, and silently skipped the entire replacement block when no such prefix was found.

This left stale `NODE_{old_hostname}_*` references in the container's command args while the env var definitions were correctly renamed, causing etcd to fail at startup with "environment variable `NODE_<old>_ETCD_URL_HOST` must be set".

Apply hostname replacements to all command args unconditionally instead of gating on the shebang heuristic.


danmanor (approved):

omertuc:
Overriding cause we're looking better than https://github.com/rh-ecosystem-edge/recert/pull/1636

PR #1757, opened 2026-07-08
Preserve IngressController during cluster domain rename

jmontesi:
Instead of deleting the default IngressController CR from etcd and letting the Ingress Operator recreate it (triggering a full router rollout and cert reconciliation cascade), preserve the IC and fix it in-place:

- Remove IC deletion from delete_resources()
- Add fix_ingresscontroller_status_domain() to update the IC's status.domain to the new cluster domain

This eliminates ~1m47s of CO convergence time post-upgrade by avoiding the IC deletion/recreation cascade and the cert synchronization delay.

Ref: [CNF-22166](https://redhat.atlassian.net/browse/CNF-22166)

### DIFF
--- a/src/ocp_postprocess/cluster_domain_rename/etcd_rename.rs
+++ b/src/ocp_postprocess/cluster_domain_rename/etcd_rename.rs
@@ -44,11 +44,6 @@ pub(crate) async fn delete_resources(etcd_client: &Arc<InMemoryK8sEtcd>) -> Resu
                     .await?,
             )
             .chain(etcd_client.list_keys("apiserver.openshift.io/apirequestcounts/").await?)
-            .chain(
-                etcd_client
-                    .list_keys("operator.openshift.io/ingresscontrollers/openshift-ingress-operator/default")
-                    .await?,
-            )
             .map(|key| async move {
                 etcd_client.delete(&key).await.context(format!("deleting {}", key))?;
                 Ok(())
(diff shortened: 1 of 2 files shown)


src/ocp_postprocess/cluster_domain_rename/etcd_rename.rs
```diff
@@ -159,6 +155,108 @@ pub(crate) async fn fix_router_certs(
...
+
+    let tls_crt_val = router_data.get("tls.crt").context("no tls.crt in router-certs-default")?;
+    let tls_key_val = router_data.get("tls.key").context("no tls.key in router-certs-default")?;
+
+    let cert_pem = value_to_bytes(tls_crt_val).context("reading tls.crt bytes")?;
+    let key_pem = value_to_bytes(tls_key_val).context("reading tls.key bytes")?;
+
+    let mut combined = Vec::with_capacity(cert_pem.len() + key_pem.len() + 1);
+    combined.extend_from_slice(&cert_pem);
+    if !cert_pem.ends_with(b"\n") {
```
omertuc: Please squash commits that only fix problems introduced in your previous commits

src/ocp_postprocess/cluster_domain_rename/etcd_rename.rs
```diff
@@ -159,6 +155,108 @@ pub(crate) async fn fix_router_certs(
...
+        Value::String(BASE64.encode(bytes))
+    } else {
+        Value::Array(bytes.iter().map(|b| Value::Number((*b as u64).into())).collect())
+    }
+}
+
+/// Copy the wildcard cert+key from router-certs-default (openshift-ingress) into
+/// v4-0-config-system-router-certs (openshift-authentication) so that the OAuth
+/// server has the same cert the router is serving. Without this, recert's crypto
+/// scan regenerates independent certs for each secret, creating a mismatch that
```
omertuc: >Add sync_router_cert_to_auth_secret() to copy the router wildcard cert into the authentication operator's v4-0-config-system-router-certs secret

recert doesn't regenerate independent certs, it makes sure to keep them the same everywhere they were already the same, so this change, or at least the justification given for it, doesn't make sense for me. Can you please provide concrete examples where you saw this difference, and provide a technical explanation for why the main generic recert code didn't handle it?

omertuc (changes requested):
Thanks for the PR. A couple of things:

src/ocp_postprocess/cluster_domain_rename/etcd_rename.rs
jmontesi: Yes, that's probably not needed. When I first removed the IC deletion, the upgrade failed with the authentication operator permanently degraded (v4-0-config-system-router-certs missing/stale). I added both `fix_ingresscontroller_status_domain()` and `sync_router_cert_to_auth_secret()` together as the fix, and validated that the combined change worked across several upgrade cycles.

However, I didn't test the two fixes independently. Given that recert's generic code handles cert identity, it's likely that only `fix_ingresscontroller_status_domain()` is actually needed (the stale status.domain causes the Ingress Operator to unpublish/regenerate certs for the wrong domain, breaking the auth chain). I'll remove `sync_router_cert_to_auth_secret()` and re-test to confirm.

src/ocp_postprocess/cluster_domain_rename/etcd_rename.rs
jmontesi: Will do.

src/ocp_postprocess/cluster_domain_rename/etcd_rename.rs
omertuc: Alright :+1:

omertuc:

omertuc:

omertuc:

PR #1758, opened 2026-07-09
feat: support ECDSA private keys in PKCS#8 and SEC1 formats

sebrandon1:
## Summary

Two gaps in ECDSA key format handling:

1. **Scanning path** (`process_pem_private_key`): bailed on ECDSA keys in PKCS#8 format (`"PRIVATE KEY"` tag) even though recert fully supports ECDSA through the SEC1 path. Fixed by extracting PKCS#8 DER bytes and deriving the public key via `openssl pkey -inform DER -pubout` (curve-agnostic).

2. **`--use-key` path** (`key_from_pem`): bailed on SEC1 EC keys (`"EC PRIVATE KEY"` tag) even though the scanning path already converts SEC1 → PKCS#8 via `openssl pkcs8 -topk8 -nocrypt`. Applied the same conversion so `--use-key` rules accept EC keys in either format.

## Context

Fixes [OCPBUGS-94076](https://issues.redhat.com/browse/OCPBUGS-94076). Supersedes [lifecycle-agent#7610](https://github.com/openshift-kni/lifecycle-agent/pull/7610) (closed) — handling the format conversion in recert directly is cleaner than normalizing in LCA.

Related: [recert#1739](https://github.com/rh-ecosystem-edge/recert/issues/1739) (Ed25519 support — separate effort)

## Design decisions

- **`openssl pkey -inform DER -pubout`** for public key extraction in scanning — curve-agnostic, consistent with existing openssl patterns
- **`openssl pkcs8 -topk8 -nocrypt`** for SEC1 → PKCS#8 conversion in `--use-key` — mirrors `process_pem_ec_private_key()`
- **PEM-encoded SPKI bytes in `PublicKey::Ec`** — matches `from_ec_cert_bytes()` for correct `Hash`/`Eq` cert-to-key pairing
- **`Bytes::from(pkcs8_der)`** — zero-copy move of the owned Vec

## Unit tests added

| Test | File | Verifies |
|------|------|----------|
| P-256 ECDSA PKCS#8 | `crypto_objects.rs` | Key accepted, stored as `PrivateKey::Ec` with original DER bytes, public key is valid PEM |
| P-384 ECDSA PKCS#8 | `crypto_objects.rs` | Same for P-384 — confirms curve-agnostic handling |
| RSA PKCS#8 regression | `crypto_objects.rs` | RSA keys still route to `PrivateKey::Rsa` |
[...]

PR #1827, opened 2026-07-21
fix: use correct PRIVATE KEY PEM tag for ECDSA PKCS#8 keys

sebrandon1:
## Summary

Corrects the PEM tag used when serializing ECDSA PKCS#8 private keys. Since #1758, PrivateKey::Ec stores PKCS#8 DER bytes but the output paths emitted them with the "EC PRIVATE KEY" PEM tag (SEC1 format, RFC 5915). The correct tag for PKCS#8 is "PRIVATE KEY" (RFC 5958).

**Impact**: Low. Real-world TLS consumers (OpenSSL, Go crypto/tls, cert-manager) parse the ASN.1 content rather than relying on the PEM tag, so this does not cause functional failures. E2E IBU testing with cert-manager ECDSA certs passed with the old tag. This is a standards-compliance fix to eliminate a class of potential issues with strict PEM parsers.

## Changes

- Changed PEM tag from "EC PRIVATE KEY" to "PRIVATE KEY" in Serialize, pem(), and commit_filesystem_private_key()
- Added round-trip PEM tag assertion to existing ECDSA PKCS#8 tests
- Added 4 new tests covering SEC1 scanning, full round-trip, Serialize output, and corrupted DER

## Unit tests

| Test | File | Verifies |
|------|------|----------|
| test_process_pem_private_key_ecdsa_p256_pkcs8 | crypto_objects.rs | P-256 PKCS#8 round-trip + PEM tag assertion |
| test_process_pem_private_key_ecdsa_p384_pkcs8 | crypto_objects.rs | P-384 PKCS#8 round-trip + PEM tag assertion |
| test_process_pem_ec_private_key_sec1_p256 | crypto_objects.rs | SEC1 scanning path produces PrivateKey::Ec + valid PublicKey |
| test_ec_pkcs8_full_round_trip | crypto_objects.rs | pem() output re-parsed by process_single_pem succeeds |
| test_ec_serialize_uses_correct_pem_tag | crypto_objects.rs | Serialize impl emits "PRIVATE KEY" tag, not "EC PRIVATE KEY" |
| test_process_pem_private_key_corrupted_pkcs8 | crypto_objects.rs | Corrupted PKCS#8 DER rejected with error |
| test_process_pem_private_key_rsa_pkcs8_still_works | crypto_objects.rs | RSA PKCS#8 regression guard |

## Files changed

| File | Change |
|------|--------|
| keys.rs | Serialize and pem() -- tag fix |
| distributed_private_key.rs | commit_filesystem_private_key() -- same tag fix |
| crypto_objects.rs | PEM tag round-trip assertions + 4 new tests |

## Context

Follow-up to #1758 which correctly changed the input path to store PKCS#8 DER in PrivateKey::Ec, but did not update the output paths to use the matching PEM tag.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved elliptic-curve private-key handling and serialization.
  * Standardized EC private-key output to use the `PRIVATE KEY` PEM format.
  * Preserved key data during PEM round trips and when storing keys on disk or in Kubernetes.
  * Improved handling of corrupted private-key data and unsupported PEM labels.
  * Ensured regenerated keys are committed consistently across supported storage methods.

* **Tests**
  * Expanded coverage for EC key formats, serialization, round trips, and error scenarios.


sebrandon1:
Hey @omertuc, could you mark these ok-to-test when you get a chance? All builds are passing on GitHub Actions. Thanks!

sebrandon1:
The `baremetalds-sno-recert-cluster-rename` CI failure on this PR is a known infrastructure issue — the test step polls the node at its original IP after recert changes it, causing SSH timeouts. A fix is in progress: https://github.com/openshift/release/pull/82217

sebrandon1:
https://github.com/openshift/release/pull/82217 is now merged — issuing retests to see if it passes.

/retest

omertuc:
Comment authored by Claude after a conversation about something that bothered me a bit that I think we should improve:
  
  
>The tag fix is correct — PKCS#8 DER should use `PRIVATE KEY`, not `EC PRIVATE KEY`. But this means if the original file on disk was SEC1 (`EC PRIVATE KEY`), recert now silently converts it to PKCS#8 on output. The format conversion already happens at scan time in `process_pem_ec_private_key` (openssl pkcs8 -topk8), and the original encoding is lost at that point.
>
>For correctness, we should preserve the original encoding and write it back in the same format. Either track it in the variant (`Ec { der: Bytes, encoding: EcEncoding }`) or split into `EcSec1` / `EcPkcs8`. The PKCS#8 conversion can still happen internally for key operations, but the output format should match the input.



Not a blocker for this PR since the bytes were already being converted to PKCS#8 before, but if we're already fixing it, it would be nice to do it proper

sebrandon1:
I can take a look at that.

sebrandon1:
Addressed the encoding feedback from https://github.com/rh-ecosystem-edge/recert/pull/1827#issuecomment-5762380619.

ECDSA keys now keep the PEM form they were scanned with:

- PKCS#8 (`PRIVATE KEY`) stays PKCS#8
- SEC1 (`EC PRIVATE KEY`) is still converted to PKCS#8 for key ops, but is written back as SEC1

Storage is `PrivateKey::Ec { der, encoding }` where `der` is always PKCS#8 and `encoding` only affects output. Equality and hashing ignore encoding so the same key still merges if it shows up in both forms.

Also covered:

- Regenerated keys copy the scanned encoding
- Unit tests for PKCS#8 and SEC1 round-trips, serialize tags, conversion helpers, and encoding copy
- Integration checks: PKCS#8 fixtures stay `PRIVATE KEY`; P-256/P-384 scenarios force SEC1 input and assert regenerated keys stay `EC PRIVATE KEY`

omertuc:
I've asked Claude to review a particular aspect of the encoding marker implementation, it had this to say:

> **Main point: hide the encoding inside an `EcPrivateKey` type**
>
> `PrivateKey::Ec` stores the encoding, but its fields are public, so every caller has to deal with `EcEncoding`. That's why it shows up in signatures and call sites across several files. Suggest:
>
> ```rust
> enum EcEncoding { Sec1, Pkcs8 }  // private to keys.rs
>
> #[derive(Clone)]
> pub(crate) struct EcPrivateKey { pkcs8_der: Bytes, encoding: EcEncoding }
> // PartialEq/Eq/Hash on pkcs8_der only
>
> impl EcPrivateKey {
>     pub(crate) fn from_pkcs8_der(der: Bytes) -> Self;
>     pub(crate) fn from_sec1_pem(pem: &pem::Pem) -> Result<Self>;  // does the openssl conversion
>     pub(crate) fn pkcs8_der(&self) -> &Bytes;
>     pub(crate) fn pem(&self) -> Result<pem::Pem>;
> }
>
> pub(crate) enum PrivateKey { Rsa(..), Ec(EcPrivateKey), Ed25519(..) }
> ```
>
> Specific comments:
>
> - **`keys.rs`: the custom `PartialEq`/`Hash` impls.** These rebuild equality and hashing for the whole `PrivateKey` enum just to skip one field. If they move onto `EcPrivateKey`, `PrivateKey` can go back to `#[derive(Hash, Eq, PartialEq)]`, and the `mem::discriminant` code goes away.
> - **`keys.rs`: `ec_encoding()`, `with_ec_encoding(Option<EcEncoding>)` and `with_ec_encoding_from()`.** That's three public functions for one job: "the new key should be written like the old one." One `with_encoding_of(self, original: &PrivateKey)` is enough, and it keeps `EcEncoding` out of the public API.
> - **`crypto_objects.rs`: `process_pem_private_key_encoded`.** The extra function exists only to pass the encoding through. With `EcPrivateKey::from_sec1_pem`, `process_pem_ec_private_key` becomes:
>   ```rust
>   let private_part = PrivateKey::Ec(EcPrivateKey::from_sec1_pem(pem)?);
>   let public_part = PublicKey::try_from(&private_part)?;
>   ```
>   and `process_pem_private_key` goes back to its single original form.
> - **`cert_key_pair.rs:106-107`.** This borrows the key, reads its encoding, then borrows it mutably to set it. The same logic also appears in `DistributedPrivateKey::regenerate`, written differently. Suggest a `DistributedPrivateKey::set_regenerated(&mut self, key: PrivateKey)` that applies the original's encoding, and have both places call it. Callers then can't forget to preserve the encoding.
> - **`distributed_private_key.rs`: `commit_at_location`, `commit_k8s_private_key` and `commit_filesystem_private_key` now take `private_key_pem`.** I guess this was done so a SEC1 key only goes through openssl once. That's reasonable, but it adds a parameter to three functions. Unless there are a lot of locations, calling `pem()` inside each function as before seems fine. Otherwise, it could be cached on the key.
> - **Tests.** Many tests match on `Ec { der, encoding }` and assert on `EcEncoding` values, so they depend on the internal layout. Checking the output instead is sturdier and still works once the fields are private: assert that `pem().tag()` is `"PRIVATE KEY"` or `"EC PRIVATE KEY"`, and compare keys with `==`. The two `with_ec_encoding_from` tests and the identity/hash test would fit better in a `keys.rs` test module.
>
> One more comment, not about encapsulation:
>
> - **`crypto_objects.rs`: `process_pem_ec_private_key`.** It now goes through `signing_key_pair_from_pkcs8_der`, where it used to only extract the public key. That extra validation is fine; it's just a quiet change in behaviour that could go in the PR description.

sebrandon1:
Thanks for the detailed feedback — all of it is addressed in the latest push.

**`EcPrivateKey` newtype + private `EcEncoding`** — Done. `EcEncoding` is now private to `keys.rs`; callers only interact with `EcPrivateKey::from_pkcs8_der`, `from_sec1_pem`, `pkcs8_der()`, and `pem()`.

**`PartialEq`/`Hash` on `EcPrivateKey`** — `PartialEq`/`Eq`/`Hash` are implemented directly on `EcPrivateKey` (over `pkcs8_der` only, ignoring encoding). `PrivateKey`'s impls now just delegate to the per-variant impls. (`PrivateKey` still has explicit impls rather than `#[derive]` because `RsaPrivateKey`'s hash doesn't include `mem::discriminant` automatically, but the custom `EcEncoding`-aware logic is fully in `EcPrivateKey` now.)

**Three encoding helpers → `with_encoding_of`** — Done. `ec_encoding()`, `with_ec_encoding(Option<EcEncoding>)`, and `with_ec_encoding_from()` are gone; `with_encoding_of(self, original: &PrivateKey)` does the job.

**`process_pem_private_key_encoded` shim** — Removed. `process_pem_ec_private_key` is now:
```rust
let private_part = PrivateKey::Ec(EcPrivateKey::from_sec1_pem(pem)?);
let public_part = PublicKey::try_from(&private_part)?;
```

**`cert_key_pair.rs:106-107`** — Cleaned up using `with_encoding_of` directly. Went with a direct call rather than a dedicated `set_regenerated` helper, since the two call sites are structurally different enough that a helper wouldn't simplify both cleanly.

**`private_key_pem` parameter** — Reverted. Each commit function calls `self.key_regenerated.as_ref().context(...)?.pem()?` inline, as you suggested.

**Tests** — Moved the three encoding-focused tests to `keys.rs`. All tests now check `pem().tag()` (`"PRIVATE KEY"` or `"EC PRIVATE KEY"`) and compare keys with `==` rather than matching on internal `EcEncoding` values.

**`process_pem_ec_private_key` now validates via `signing_key_pair_from_pkcs8_der`** — Good catch on flagging this. It does add an extra key-parse step on SEC1 input. Added a note to the PR description.

src/cluster_crypto/crypto_objects.rs
```diff
@@ -169,25 +169,19 @@ fn process_pem_private_key(pem: &pem::Pem) -> Result<Option<CryptoObject>> {
...
             Some((private_part, public_part).into())
         }
         InMemorySigningKeyPair::Rsa(_, bytes) => {
             let rsa_private_key = rsa::RsaPrivateKey::from_pkcs1_der(&bytes)?;
-
             let private_part = PrivateKey::Rsa(rsa_private_key);
             let public_part = PublicKey::try_from(&private_part)?;
-
             Some((private_part, public_part).into())
         }
```
omertuc: This match has too much repetitive code now. Maybe move just the private creation part into the match, the public building and tuple building seems shared across all

sebrandon1: I'll take a look.

sebrandon1: Done. The match in `process_pem_private_key` now only builds the private key; public key derivation and tuple building are shared. I also swept the rest of the PR for the same pattern and deduped the JWT verify/resign openssl calls (`digest_for_alg`), `sign`, the EC curve OID parsing, and the repeated test key generation (new `test_utils`).

